Issue #31, opened 2024-02-13
Add proper validation error display for the initial registration form


wintermeyer:
The main user registration form doesn't display any validation errors. Any validation error leads to a redirect to `http://localhost:4000/auth/user/password/register` which doesn't make sense. 

<img width="373" alt="Bildschirmfoto 2024-02-13 um 16 51 50" src="https://github.com/animina-dating/animina/assets/69770/2f38e89e-ec72-4db8-9b0e-a2911738c5f3">

# Tasks

- [ ] Make sure that a validation error at http://localhost:4000 will show the same form (same URL) again but with additional information about the validation errors. Preferably as small red text below the fields which need to be fixed.

wintermeyer:
https://github.com/animina-dating/animina/pull/38

Issue #33, opened 2024-02-13
Display validation errors for the second step of the registration


wintermeyer:
http://localhost:4000/registration/potential-partner doesn't display validation errors. Please fix that.

See:
- https://github.com/animina-dating/animina/issues/31
- https://github.com/animina-dating/animina/issues/32

wintermeyer:
https://github.com/animina-dating/animina/pull/38

Issue #32, opened 2024-02-13
Add validation for age and height


wintermeyer:
The `User` resource has the following attributes

```elixir
attribute :minimum_partner_height, :integer, allow_nil?: true
attribute :maximum_partner_height, :integer, allow_nil?: true
attribute :minimum_partner_age, :integer, allow_nil?: true
attribute :maximum_partner_age, :integer, allow_nil?: true
```

Please add validations that the the max number has to be the same or higher than the min number and vice versa.

wintermeyer:
https://github.com/animina-dating/animina/pull/38

PR #38, opened 2024-02-14
show validation errors on basic form inputs

briankariuki:
This commit adds form input validations, fixes issue #31, #32, #33

### DIFF
--- a/lib/animina/accounts/resources/user.ex
+++ b/lib/animina/accounts/resources/user.ex
@@ -61,6 +61,13 @@ defmodule Animina.Accounts.User do
     prepare build(load: [:gravatar_hash, :age, :credit_points])
   end
 
+  validations do
+    validate {Animina.Validations.MinMaxAge, attribute: :maximum_partner_age}
+    validate {Animina.Validations.MinMaxAge, attribute: :minimum_partner_age}
+    validate {Animina.Validations.MinMaxHeight, attribute: :minimum_partner_height}
+    validate {Animina.Validations.MinMaxHeight, attribute: :maximum_partner_height}
+  end
+
   authentication do
     api Animina.Accounts
 
--- a/lib/animina/validations/min_max_age.ex
+++ b/lib/animina/validations/min_max_age.ex
@@ -7,10 +7,9 @@ defmodule Animina.Validations.MinMaxAge do
 
   @impl true
   def init(opts) do
-    if is_atom(opts[:attribute]) do
-      {:ok, opts}
-    else
-      {:error, "attribute must be an atom!"}
+    case is_atom(opts[:attribute]) do
+      true -> {:ok, opts}
+      _ -> {:error, "attribute must be an atom!"}
     end
   end
 
@@ -20,6 +19,9 @@ defmodule Animina.Validations.MinMaxAge do
     max_age = Ash.Changeset.get_attribute(changeset, :maximum_partner_age)
 
     cond do
+      min_age == nil || max_age == nil ->
+        :ok
+
       opts[:attribute] == :maximum_partner_age && max_age < min_age ->
         {:error,
          field: opts[:attribute], message: "must be same or higher than minimum partner age"}
--- a/lib/animina/validations/min_max_age.ex
+++ b/lib/animina/validations/min_max_age.ex
@@ -1,0 +1,35 @@
+defmodule Animina.Validations.MinMaxAge do
+  use Ash.Resource.Validation
+
+  @moduledoc """
+  This is a module for validating the mimimum and maximum partner ages
+  """
+
+  @impl true
+  def init(opts) do
+    if is_atom(opts[:attribute]) do
+      {:ok, opts}
+    else
+      {:error, "attribute must be an atom!"}
+    end
+  end
+
+  @impl true
+  def validate(changeset, opts) do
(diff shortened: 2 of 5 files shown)


lib/animina/accounts/resources/user.ex
wintermeyer, line 1: Maybe a miscommunication from my side. The user should be able to use "doesn't matter". Please add that case as a valid possibility to the validations.

Bildschirmfoto 2024-02-15 um 07 51 41

lib/animina/validations/min_max_age.ex
```diff
@@ -0,0 +1,35 @@
+defmodule Animina.Validations.MinMaxAge do
+  use Ash.Resource.Validation
+
+  @moduledoc """
+  This is a module for validating the mimimum and maximum partner ages
+  """
+
+  @impl true
+  def init(opts) do
+    if is_atom(opts[:attribute]) do
```
wintermeyer: In Elixir I prefer a `case` over an `if`. Especially in cases when there is a chance that more stuff gets added in the future. Please change this to a `case` statement.

BTW: An `if` becomes a `case` under the hood. https://hexdocs.pm/elixir/case-cond-and-if.html

Bildschirmfoto 2024-02-15 um 07 55 24

lib/animina/validations/min_max_height.ex
```diff
@@ -0,0 +1,35 @@
+defmodule Animina.Validations.MinMaxHeight do
+  use Ash.Resource.Validation
+
+  @moduledoc """
+  This is a module for validating the mimimum and maximum partner heights
+  """
+
+  @impl true
+  def init(opts) do
+    if is_atom(opts[:attribute]) do
```
wintermeyer: Please use a `case` here.

lib/animina_web/live/root_live.ex
```diff
@@ -71,21 +92,42 @@ defmodule AniminaWeb.RootLive do
...
+              f,
+              :username,
               class:
-                "block w-full rounded-md border-0 py-1.5 text-gray-900 shadow-sm ring-1 ring-inset ring-gray-300 placeholder:text-gray-400 focus:ring-2 focus:ring-inset focus:ring-indigo-600 sm:text-sm sm:leading-6",
+                "block w-full rounded-md border-0 py-1.5 text-gray-900 shadow-sm ring-1 ring-inset placeholder:text-gray-400 focus:ring-2 focus:ring-inset sm:text-sm  phx-no-feedback:ring-gray-300 phx-no-feedback:focus:ring-indigo-600 sm:leading-6 " <>
+                  if(get_field_errors(f[:username], :username) != [],
+                    do: "ring-red-600 focus:ring-red-600",
+                    else: "ring-gray-300 focus:ring-indigo-600"
+                  ),
               placeholder: "Pusteblume1977",
```
wintermeyer: This should be a gettext("Pusteblume1977"). My bad. Please fix it while working here.

wintermeyer (changes requested):
@briankariuki I like your code and the way you tackled the problem! Great work!

Requests:
- Please add screenshots to the PRs in case a GUI element is involved. So that I get a quick idea when first seeing the PR without having to pull the code.
- Please use one PR per page/problem. This is ok because it is one workflow but I would have splitted it up in two PRs.

lib/animina_web/live/root_live.ex
```diff
@@ -71,21 +92,39 @@ defmodule AniminaWeb.RootLive do
...
           </label>
           <div class="mt-2">
-            <%= text_input(f, :username,
+            <%= text_input(
+              f,
+              :username,
               class:
-                "block w-full rounded-md border-0 py-1.5 text-gray-900 shadow-sm ring-1 ring-inset ring-gray-300 placeholder:text-gray-400 focus:ring-2 focus:ring-inset focus:ring-indigo-600 sm:text-sm sm:leading-6",
+                "block w-full rounded-md border-0 py-1.5 text-gray-900 shadow-sm ring-1 ring-inset  placeholder:text-gray-400 focus:ring-2 focus:ring-inset  sm:text-sm sm:leading-6 " <>
+                  if(get_field_errors(f[:username], :username) != [],
```
wintermeyer: I haven't played with this but I think `if(get_field_errors(f[:username], :username) != []` can be done better with something like `unless(get_field_errors(f[:username], :username)`

```elixir
iex(1)> if [] do
...(1)> IO.puts "a"
...(1)> else
...(1)> IO.puts "b"
...(1)> end
a
:ok
``

lib/animina/accounts/resources/user.ex
briankariuki: The validation now works when the user selects the `doesn't matter` option.

image

lib/animina/validations/min_max_age.ex
briankariuki: This has been replaced with case.

lib/animina_web/live/root_live.ex
briankariuki: I couldn't get unless to work like you suggested `unless(get_field_errors(f[:username], :username))` , instead I've opted to include `== []`  in the check.  `unless(get_field_errors(f[:username], :username) == [])`

wintermeyer (approved):

PR #41, opened 2024-02-15
add user language detection, closes #34, closes #35

### LINKED ISSUE #34: Auto lang detection
Read the accepted language from the browser header. Either it is German (something wie a `de`) than the locale will be `de` otherwise it will be `en`.  Please have a look at https://github.com/smeevil/set_locale if that makes sense go for it otherwise use a different solution. Right now we only need the Accept-Language HTTP header variant.

### LINKED ISSUE #35: Add language to the User resource
Add an `language` attribute to the `User` and `BasicUser` resources. If should save the value of the Accept-Language HTTP header (see https://github.com/animina-dating/animina/issues/34) when a new `BasicUser` is registered.

It should save the real value. So if it is `fr` it should save `fr` and not `en` even if we don't service `fr` yet.

briankariuki:
- Added a plug that extracts the `accept-language` header and adds it to the user's session params
- Added a new field to the basic user's resource to store the language, generated migrations for this.
- Added the language field to the basic user's form.


lib/animina/accounts/resources/basic_user.ex
```diff
@@ -33,6 +33,7 @@ defmodule Animina.Accounts.BasicUser do
     attribute :gender, :string, allow_nil?: false
     attribute :height, :integer, allow_nil?: false
     attribute :mobile_phone, :string, allow_nil?: false
+    attribute :language, :string, allow_nil?: false
```
wintermeyer, line 36: The `User` resource needs the `:language` attribute too.

lib/animina_web/plugs/language.ex
```diff
@@ -0,0 +1,57 @@
...
+  """
+
+  def init(options), do: options
+
+  def call(conn, _options) do
+    conn
+    |> extract_accept_language()
+    |> case do
+      [language | _] ->
+        conn |> put_session("language", String.split(language, "-") |> Enum.at(0, "en"))
```
wintermeyer: I didn't dive into the code but I am missing the `de` part here. Is it missing?

wintermeyer (changes requested):
I didn't do a good job in describing the feature request. Let me retry: I want a table entry `language` which stores the original and complete accept-language string. So if e.g. somebody has `fr-CH` I want that complete string to be stored. But because we don't provide French as a language yet I want the webpage to be displayed in English (`en`). Only in case `de-whatever` is stored I want it to be in German (`de`).

Why storing the complete string? That will help us in the future to analyse which additional languages we need to provide. The complete string gives us the maximum amount of information.

lib/animina/accounts/resources/basic_user.ex
briankariuki: I've added the language attribute to the User resource

lib/animina_web/plugs/language.ex
briankariuki: I've added some more logic to configure Gettext with the user's language
- `en-whatever` will be parsed to `en`
- `de-whatever` will be parsed to `de`

original `accept-language` header will be saved as is.

wintermeyer (approved):
I think there is room for improvement in `AniminaWeb.Plugs.AcceptLanguage` in the sense to make the code easier to understand and to maintain (e.g. adding an other language). BUT it works and does the job. No need to invest an other couple of hours just to make it nicer.

Issue #40, opened 2024-02-15
Deactivate button when invalid form data


wintermeyer:
<img width="483" alt="Bildschirmfoto 2024-02-15 um 13 10 27" src="https://github.com/animina-dating/animina/assets/69770/2e2b49b9-b205-4898-805a-85bda9f63252">

Since we know when the form data is valid we should use Tailwind CSS to deactivate the "Register new account" button until the form data is valid.

philipp-kempgen:
See `/demo` page.

`<form class="group" …>` and:
https://github.com/animina-dating/animina/blob/897fb9bb1fa1623d276ce3b79c2a4cc559771676/lib/animina_web/controllers/page_html/demo.html.heex#L380

Issue #50, opened 2024-02-16
Validate min-max height


wintermeyer:
It is very unlikely but let's validate that the height is between 50 cm and 250 cm. The validation doesn't harm.

Issue #49, opened 2024-02-16
Validate unique mobile phone number


wintermeyer:
See #48. Same problem.

philipp-kempgen:
The "Mobile phone number" input field lets me enter "abcde".
The validation should probably make sure, that this field is a valid phone number, i.e. heavily restrict the allowed characters, and then *normalize* the phone number.

Normalization is necessary so we don't end up with "0123-456789", "0123 456789", "0123456789" and "+49123456789", which is essentially the same phone number.

We also need to make sure not to bombard emergency numbers etc. with text messages.
And we probably don't want to send text messages to expensive foreign numbers.

Realistically, we need to use some API here to send text messages.
The provider would be in Germany for now.

I.e. for now I would allow `/\A[0-9+ \-]+\z/`, then (server side?):
- strip "boring" stuff, i.e. strip "`(0)`", then strip "` `" and "`-`",
- then normalize the phone number by converting leading "`00`" to "`+`",
- then strip a leading "`0`" if any,
- then add country prefix "`+49`" if the phone number doesn't start in "`+`",
- then (server side) validate that the phone number is `/\A(?<country_code>\+49)[1][0-9]{9,13}\z/`

I hope I didn't forget anything.

This may also need some rate-limiting mechanism.

And the second step should ask for a 6-digit code received by text message.

And it needs a way to trigger a retransmission, as text messages may get lost.

We may even want to remove the "verification code" thing for now, because this isn't an easy thing to get right.

philipp-kempgen:
Oh, and it would need to work in **development mode** without actually sending text messages to a cell phone number and by allowing e.g. "000000" as verification.

Not a trivial thing, overall.

wintermeyer:
> Oh, and it would need to work in **development mode** without actually sending text messages to a cell phone number and by allowing e.g. "000000" as verification.

Right now we do not send any verification text messages. That is going to be a future feature.

philipp-kempgen:
> Right now we do not send any verification text messages. That is going to be a future feature.

I know. But the input field says „Mobile phone number (to receive a verification code)“, which – to me – makes it seem like I should receive a verification code via text message during the registration.

That's why I was proposing to either actually implement the feature (and also think about development mode), or to change the wording as not to confuse the user, or to remove that feature for now and implement 2FA later.

wintermeyer:
[...]

Issue #48, opened 2024-02-16
Validate unique email address


wintermeyer:
Email addresses have to be unique.

A validation error should be displayed:

<img width="464" alt="Bildschirmfoto 2024-02-16 um 13 29 02" src="https://github.com/animina-dating/animina/assets/69770/638dff0e-5fe5-427b-b296-86b043aad69d">

PS: Yes, this is a tricky problem because one could use this validation to check if a user has an account in the system. I have no idea how to solve this otherwise. We need a unique email address.

philipp-kempgen:
Note that URIs such as e-mail addresses should go through a library that handles normalization, as e.g. "`Bücher.example`" and "`xn--bcher-kva.example`" is the same domain name, semantically, and we probably don't want to end up with 2 users
- "`user@Bücher.example`" and
- "`user@xn--bcher-kva.example`"

(<https://en.wikipedia.org/wiki/Internationalized_domain_name>)

Granted, if someone wanted multiple accounts, getting more e-mail addresses is an easy thing (think "+" notation, think disposable e-mail addresses).

But normalization would likely still be required for non-ASCII addresses. (Or make sure they are ASCII, for now.)

wintermeyer:
@philipp-kempgen I see your point but I'd like to kick this can a bit more down the road. We can tackle that later. 

PS: https://github.com/kdeldycke/awesome-falsehood#emails

PR #51, opened 2024-02-19
adds validations for email, username, phonenumber, birthday, height, fixes issues #40,#44,#45,#46,#47,#48,#49,#50

briankariuki:
Added unique validation for username, email address and validation for minimum age
image

Added unique validation for mobile number, validation for the min, max height, min length for postal code and disables the submit form button if form is invalid.
image


lib/animina/validations/birthday.ex
```diff
@@ -0,0 +1,24 @@
...
+  @impl true
+  def validate(changeset, opts) do
+    birthday = Ash.Changeset.get_attribute(changeset, :birthday)
+    today = Date.utc_today()
+
+    cond do
+      birthday == nil ->
+        :ok
+
+      today.year - birthday.year < 18 ->
```
wintermeyer, line 25: This doesn't work for birthdays which are later in the year than today.

lib/animina/accounts/resources/basic_user.ex
```diff
@@ -29,9 +29,21 @@ defmodule Animina.Accounts.BasicUser do
     end
 
     attribute :birthday, :date, allow_nil?: false
-    attribute :zip_code, :string, allow_nil?: false
+
+    attribute :zip_code, :string do
+      constraints max_length: 5,
```
wintermeyer: I am not a big fan of the validation error here. "Postal code length must be greater than or equal to 5" is not correct. It should be something like "Postal code length must be 5".

Either we have to "fix" the validation error or we have to find a solution to use some sort of `length: 5`

lib/animina/accounts/resources/basic_user.ex
```diff
@@ -91,8 +107,9 @@ defmodule Animina.Accounts.BasicUser do
   end
 
   identities do
-    identity :unique_email, [:email]
-    identity :unique_username, [:username]
+    identity :unique_email, [:email], eager_check_with: Animina.Accounts
```
wintermeyer: Wouldn't it improve code readability if we'd use an alias for `Animina.Accounts`?

lib/animina/accounts/resources/user.ex
```diff
@@ -67,6 +79,7 @@ defmodule Animina.Accounts.User do
     validate {Animina.Validations.MinMaxAge, attribute: :minimum_partner_age}
```
wintermeyer: Wouldn't it improve code readability if we'd use an alias for `Validations`?

wintermeyer (changes requested):

lib/animina/validations/birthday.ex
briankariuki: This is fixed

lib/animina/accounts/resources/basic_user.ex
briankariuki: I've added a custom validation for the zip_code attribute, since Ash.Type.String constraints does not support a length validation.

wintermeyer (approved):

PR #60, opened 2024-02-23
Feature/input fixes (closes #58, closes #55, closes #46)

### LINKED ISSUE #46: Add validation for zip codes (5 digits)
In the current version of the application all zip codes are in Germany. German zip codes always have 5 digits. They can start with a 0 but they always have to be 5 digits long.

Please display a validation error:

Bildschirmfoto 2024-02-16 um 13 24 46

### LINKED ISSUE #55: Validate gender
Validate the gender selection on the server.
Just because it is restricted on the client by the radiobox values doesn't mean someone can't change these.

### LINKED ISSUE #58: Fix form labels
The
```html
  <label for="...">...</label>
```
needs to point to the corresponding
```html
  <input id="..." .../>
```
so you can click onto a label to focus the corresponding input field.

Example: If there's an
```html
  <input id="user_username" ... />
```
then the label has to be
```html
  <label for="user_username" ... />
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

> Associating a `<label>` with a form control, such as `<input>` or `<textarea>` offers some major advantages:
> 
> - The label text is not only visually associated with its corresponding text input; it is programmatically associated with it too. This means that, for example, a screen reader will read out the label when the user is focused on the form input, making it easier for an assistive technology user to understand what data should be entered.
> - When a user clicks or touches/taps a label, the browser passes the focus to its associated input (the resulting event is also raised for the input). That increased hit area for focusing the input provides an advantage to anyone trying to activate it — including those using a touch-screen device.
> 
> To explicitly associate a `<label>` element with an `<input>` element, you first need to add the `id` attribute to the `<input>` element. Next, you add the `for` attribute to the `<label>` element, where the value of `for` is the same as the `id` in the `<input>` element.
> 
> Alternatively, you can nest the `<input>` directly inside the `<label>`, in which case the `for` and `id` attributes are not needed because the association is implicit:

philipp-kempgen:


lib/animina/accounts/resources/basic_user.ex
```diff
@@ -42,7 +42,14 @@ defmodule Animina.Accounts.BasicUser do
                   allow_empty?: false
     end
 
-    attribute :gender, :string, allow_nil?: false
+    attribute :gender, :string do
+      allow_nil? false
+      
+      constraints(
+        match: ~r/\A [mfx] \z/x,
```
wintermeyer, line 49: In case there are no technical reasons I would prefer `male`, `female` and `diverse` over `m`, `f` and `x`. Less thinking is needed while reading the code by using the full words.

lib/animina/validations/postal_code.ex
```diff
@@ -21,8 +21,8 @@ defmodule Animina.Validations.PostalCode do
       zip_code == nil ->
         :ok
 
-      String.length(zip_code) < 5 || String.length(zip_code) > 5 ->
-        {:error, field: opts[:attribute], message: "length must be 5"}
+      ! String.match?(zip_code, ~r/\A [0-9]{5} \z/x) ->
+        {:error, field: opts[:attribute], message: "must have 5 digits"}
```
wintermeyer, line 25: We need the error messages done with `gettext/1`.

wintermeyer (changes requested):
I like PRs to be about one topic. This one covers a multitude of problems. It is ok for now (i.a. for this PR) but please split it up in the future. Otherwise it is very hard to review it. Thanks!

lib/animina/accounts/resources/basic_user.ex
philipp-kempgen: If this is about personal preference, then I don't want to argue about it.

BTW: If I'm not mistaken the actual data value was "`Diverse / non-binary`" before this change.

OTOH if this is about technical reasons, then I don't see a point of using way more space in the database table and index than necessary.

I would suggest to mentally separate data from how you work with the objects. I.e. one could probably add something like this to the user model (pseudo-code):

```ex
def male?    ; gender == "m"; end
def female?  ; gender == "f"; end
def diverse? ; gender == "x"; end
```

But I think that won't even be necessary, since probably the only other place where the gender matters will be the matching algorithm, and there will likely not even be any literal "`male`"/"`female`"/"`diverse_or_non_binary`" in the code, as the algorithm will simply match data.

Actually the database column should be an `ENUM` instead of a `TEXT` field, but that's outside the scope of this issue. Also, there should be an index on the database column, or rather a combined index on e.g geolocation and gender or something, depending on the queries that the matching algorithm will run.

lib/animina/validations/postal_code.ex
philipp-kempgen: I totally agree, but none of the other validations use `gettext`(), and I simply copied from there:

https://github.com/animina-dating/animina/tree/a76efad2931542dc74fd5c530ce3b05213cf5354/lib/animina/validations

I also think that in order to translate properly we need a complete sentence in the validators.

=> #63

lib/animina/accounts/resources/basic_user.ex
wintermeyer: Not tested code:

```elixir
defmodule YourApp.Repo.Migrations.AddGenderType do
  use Ecto.Migration

  def up do
    execute "CREATE TYPE gender AS ENUM ('male', 'female', 'diverse')"
  end

  def down do
    execute "DROP TYPE gender"
  end
end
```

```elixir
defmodule YourApp.Accounts.User do
  use Ecto.Schema
  import Ecto.Changeset

  schema "users" do
    field :name, :string
    field :gender, :string, comment: "ENUM('male', 'female', 'diverse')"
    # Andere Felder...
  end

  def changeset(user, attrs) do
    user
    |> cast(attrs, [:name, :gender])
    |> validate_inclusion(:gender, ["male", "female", "diverse"])
  end
end
```

@philipp-kempgen Fire at will (which includes changing nothing at all). We do have bigger fish to fry but if you want to tackle this: Go ahead.

wintermeyer:
@philipp-kempgen What is the status of this PR? I just want to make sure that you are not waiting on something from my side which I am not aware of.

philipp-kempgen:
This code is useful already, so I would merge this stuff as quickly as possible before the code starts to rot, and then make a new enhancement issue out of the ENUM thing.
Waiting always has a risk of creating merge conflicts that may need hours to solve, especially if there are database migrations involved.

.github/workflows/elixir.yml
```diff
@@ -1,10 +1,12 @@
 name: Elixir CI
 
-on:
-  push:
-    branches: [main]
-  pull_request:
-    branches: [main]
+#on:
```
wintermeyer, line 3: What is this for?

wintermeyer (changes requested):

wintermeyer:
@philipp-kempgen I split up this PR into separate PRs. I didn't create a PR for stuff that I didn't understand.

.github/workflows/elixir.yml
philipp-kempgen: I wanted the Elixir tests to run on my feature branches, and this prevented it by restricting the action to the "`main`" branch.
Now that I think about it you may want something like:
```
on:
  push:
  pull_request:
```

PR #62, opened 2024-02-23
add uploading user photos, closes #39

### LINKED ISSUE #39: Add a user avatar
Right now we use the gravatar images for new users. That will not fly for most users. As a third step in the registration process we have to ask the user to upload or take a photo.

## Thoughts

- We need a SQUARE avatar photo of the user. If a user uploads a non square photo we have to autocrop it. Tricky. Maybe a task for AI. Worst case: We just auto crop it for now without any AI. Just do it and optimise it later.
- In case a Gravatar photo exists we should use that as a default. Saves the user some time.
- The photo has to have a minimum and a maximum resolution (size).
- We need to save the original uploaded file and a compress version (which we compress on the command line).
- In future steps we are going to ask the user for more photos which will be displayed in his/her profile. So this is going to be a `has_many` thing. 
- In the future the user has to have the possibility to change/delete those photos.
- We actually have to implement some sort of workflow to control all uploaded photos by a human team. Otherwise people will upload illegal stuff. So a new photo will be visible for the user but for everybody else only after a human has checked it.

@briankariuki In case you have a good idea: Please a) just go for it or b) discuss it here in the issue first. Feel free to create new issues and to split this one up.

briankariuki:
The photo form looks like this before a photo is selected:
image

Upon selecting a photo:
image

For photo reviews, I've implemented a simple state machine with `ash_state_machine` that looks like this. The default state is `:pending_review`:

```
  stateDiagram-v2
  pending_review --> in_review: review
  in_review --> approved: approve
  approved --> in_review: report
  approved --> error: error
  in_review --> rejected: reject
  rejected --> error: error
  in_review --> error: error
  pending_review --> error: error

```


wintermeyer (approved):

wintermeyer:
@briankariuki Where is the file stored in the file system? How is that configured?

PR #64, opened 2024-02-25
Save uploaded images to /priv/static/uploads/

briankariuki:


wintermeyer (changes requested):
I get this error:
```elixir
[error] GenServer #PID<0.2549.0> terminating
** (File.CopyError) could not copy from "/var/folders/d3/tm0bnyrx07lghmfv8851nzxr0000gn/T/plug-1708-TxON/live_view_upload-1708887977-403090739439-1" to "/Users/stefan/GitHub/animina/_build/dev/lib/animina/priv/static/uploads/9313e874-45d3-416b-80cf-e11356dc00ec.jpg": no such file or directory
```

After creating the directory by hand it works. Either add it to the README.md or create the directory by code. Or use a different directory.

The birthday field acts funny. It changes values:
birthday-problem

assets/js/app.js
```diff
@@ -16,26 +16,37 @@
...
+import { LiveSocket } from "phoenix_live_view";
+import topbar from "../vendor/topbar";
+import ImageCropper from "./hooks/cropper";
 
-let csrfToken = document.querySelector("meta[name='csrf-token']").getAttribute("content")
-let liveSocket = new LiveSocket("/live", Socket, {params: {_csrf_token: csrfToken}})
+let Hooks = {};
+
+Hooks = {
+  ImageCropper,
```
wintermeyer, line 29: When I upload a non square photo I don't see an ImageCropper anywhere. Is that a bug?

wintermeyer (changes requested):
Please correct me if I am wrong: We do a bit JavaScript magic in this PR just to realise an ImageCrop mechanism, don't we? The alternative is to autocrop the image, isn't it?

Seeing already the problems with the birthday field in the / page I think it is better to not include the ImageCropper and to keep all the JavaScript as vanilla as possible. We autocrop the images. If the user doesn't like it he/she can upload an other image.

assets/js/app.js
wintermeyer: Let's remove the ImageCropper functionality.

briankariuki:
> Seeing already the problems with the birthday field in the / page I think it is better to not include the ImageCropper and to keep all the JavaScript as vanilla as possible. We autocrop the images. If the user doesn't like it he/she can upload an other image.

The cropping here is using vanilla js by painting the image to a canvas with predefined width and height which crops it. I've fixed the bug. The cropping now works as expected. This does not affect the birthday input as the ImageCropper hook is only loaded in the upload photo screen. I've also fixed the uploads directory error.

I'm investigating why the birthday input is not working as expected.

assets/js/app.js
briankariuki: > When I upload a non square photo I don't see an ImageCropper anywhere. Is that a bug?

This is fixed. You can try with a non-square image

wintermeyer (changes requested):
It still doesn't work.

ScreenFlow

briankariuki:
Can you send me that photo file? I try it out on my end.Also does the github actions deploy the application to the web? If so can I get the link to test out the hosted environment?On 26 Feb 2024, at 10:24, Stefan Wintermeyer ***@***.***> wrote:﻿
@wintermeyer requested changes on this pull request.

It still doesn't work.
ScreenFlow.gif (view on web)

wintermeyer:
> Can you send me that photo file?

I just did by email.

> Also does the github actions deploy the application to the web?

No. Not yet. I will do that soon. I'll inform you when that happens.

BTW: I am testing on macOS with Safari.

briankariuki:
> BTW: I am testing on macOS with Safari.

I've tested on safari and I've noticed the bug. Specifically its from phoenix_live_view's live_uploader.js and not our upload/custom hook implementation.
image

It works in Chrome and I would suggest we upgrade phoenix_live_view to the latest version.

image

wintermeyer:
> I would suggest we upgrade phoenix_live_view to the latest version

Go for it.

philipp-kempgen:
> ```elixir
> [error] GenServer #PID<0.2549.0> terminating
> ** (File.CopyError) could not copy from "/var/folders/d3/tm0bnyrx07lghmfv8851nzxr0000gn/T/plug-1708-TxON/live_view_upload-1708887977-403090739439-1" to "/Users/stefan/GitHub/animina/_build/dev/lib/animina/priv/static/uploads/9313e874-45d3-416b-80cf-e11356dc00ec.jpg": no such file or directory
> ```
> 
> After creating the directory by hand it works. Either add it to the README.md or create the directory by code. Or use a different directory.

How about you add the "`priv/static/uploads`" directory to the Git repository by adding an empty "`.gitignore`" file to that directory?

Creating directories at runtime may not work if the project is mounted read-only into a Docker container.

wintermeyer:
> How about you add the "`priv/static/uploads`" directory to the Git repository by adding an empty "`.gitignore`" file to that directory?

Works for me.

philipp-kempgen:
> > How about you add the "`priv/static/uploads`" directory to the Git repository by adding an empty "`.gitignore`" file to that directory?
> 
> Works for me.

`.gitkeep` was what I meant. `.gitignore` should work just as well, though.

Rethinking this, a `.gitkeep` file doesn't make too much of a difference, because `priv/static/uploads` would still need to be mounted read-write. I guess I would still add the file to avoid having to making `priv/static` read-write, but it's not the be-all and end-all.

wintermeyer:
@briankariuki What is the status of this PR? Please request a review when you are done with your work.

briankariuki:
> @briankariuki What is the status of this PR? Please request a review when you are done with your work.

This PR is complete. You can close it.

wintermeyer:
@briankariuki Please always use the reviews functionality if a PR is good to go. Otherwise I don't know what the status is.

PR #83, opened 2024-02-27
Save phone number in e.164 format. closes #57

### LINKED ISSUE #57: Store mobile phone number in e164 format
See #49 for more information. We need to save the phone number in the [a164 format](https://en.wikipedia.org/wiki/E.164). Only that way we can be sure that mobile phone numbers are unique. But we have to display them for the human user in the human readable format.

# Tasks

- [ ] Implement https://github.com/ex-phone-number/ex_phone_number
- [ ] For now always use `{:ok, phone_number} = ExPhoneNumber.parse("...", "DE")` (DE being the important part)
- [ ] Use `ExPhoneNumber.format(phone_number, :e164)` to save the mobile phone number in the database table.
- [ ] Use `ExPhoneNumber.format(phone_number, :national)` to display the phone number in the form or anywhere else on the GUI.
- [ ] Add a validation to make sure that `ExPhoneNumber.format(phone_number, :e164)` results in a string which is at least 8 characters long. Otherwise that phone number doesn't make sense in German.

Valid phone numbers for testing:

- 0261-9886803
- 015112345678
- +4917212345678

@philipp-kempgen That is good enough for now. In case we run into edge cases in the future we will open another issue.

wintermeyer:


lib/animina/custom_types/phone_number.ex
```diff
@@ -0,0 +1,48 @@
...
+  def cast_input(nil, _), do: {:ok, nil}
+
+  def cast_input(value, _) do
+    ecto_type_cast = Ecto.Type.cast(:string, value)
+
+    case ecto_type_cast do
+      {:ok, raw_phone_number} ->
+        case ExPhoneNumber.parse(value, "DE") do
+          {:ok, phone_number} ->
+            case ExPhoneNumber.is_possible_number?(phone_number) do
```
philipp-kempgen: Is there a validation to prevent stuff like `+49 112`?

wintermeyer:
@philipp-kempgen I can see your comment in my email inbox but can't see it here. Anyhow:



IMO this is a very solid solution for the given problem.

PR #86, opened 2024-02-27
maybe fix birthday input, remove imagecropper, closes #76, closes #78

### LINKED ISSUE #76: Freak "scroll up and delete form" problem
Since a couple of days (I wanna say it is related to the ImageCropping shenanigans - but I have no proof for that) I have the following problem: I pull new code from GitHub. Do a DB reset. Fire up the Phoenix application. Visit http://localhost:4000 and start filling out the form. In the `birthday` field something strange happens: The form deletes itself and the curser jumps up to the `username` field again. I have a hard time to replicate the problem because it doesn't happen the second time. It has happened multiple times in the last 2-3 days. Current macOS and Safari.

freak-scroll

### LINKED ISSUE #78: Remove ImageCropper
I want to avoid having code in the system which we don't use and which is buggy. **Please remove all the ImageCropper related code.** We wasted already too much time on this and I want to make sure that this code is not going to become a footgun down the road. We do not need image cropping right now. Just uploading an image is good enough. We can either display just a square part of a rectangle and solve this by using CSS or we autocrop a rectangle to a square by command line tools like `convert` (ImageMagic). Both is fine for me right now.

briankariuki:
I've added a debounce on the inputs especially those that have validations, may fix the scrolling issue.


assets/js/app.js
```diff
@@ -21,14 +21,9 @@ import "phoenix_html";
 import { Socket } from "phoenix";
 import { LiveSocket } from "phoenix_live_view";
 import topbar from "../vendor/topbar";
-import ImageCropper from "./hooks/cropper";
 
 let Hooks = {};
```
wintermeyer, line 25: Is that the cleaned up version? I don't know a lot about JavaScript. I just had a look at https://github.com/animina-dating/animina/blob/1076d78d3eb7611ec7b75ef45823127bd940d358/assets/js/app.js and I want to be sure that we have a clean sweep here.

lib/animina_web/live/root_live.ex
```diff
@@ -111,7 +111,8 @@ defmodule AniminaWeb.RootLive do
               type: :text,
               required: true,
               autofocus: true,
-              autocomplete: :username
+              autocomplete: :username,
+              "phx-debounce": "300"
```
wintermeyer, line 115: Just for curiosity: What is `debounce`?

briankariuki: Debounce adds a delay for any events that get fired as the user is typing. For example validations will not happen immediately.

assets/js/app.js
briankariuki: I removed any reference to the ImageCropper hook

PR #91, opened 2024-02-28
add user interests, closes #67

### LINKED ISSUE #67: current_user defines 20 tags
With https://github.com/animina-dating/animina/pull/66 a system of tags got introduced to the application.

After uploading a photo of him/herself a user has to choose a maximum of 20 out of those tags. The GUI should display the number of already picked tags. I think it would be best to display all tags (as pills?) in a long page. Separated by the Categories. Choosing a tag should be able by clicking on it. By that its color has to change to signal that it was chosen. Once the maximum of 20 choices has been made all other tags should be displayed in a way that it makes clear that they can't be clicked on. An already clicked on tag can be unchosen by clicking again on it.

It would be good to program it in a way that we can recycle it to choose red and green flags after this step. 

# The Idea

Each user can choose red and green flags for his/her future partner.

I haven't thought about a data model yet. Please feel free to discuss it in this issue or just go ahead and implement one. Please do remember to run a `mix run priv/repo/flag_seeds.exs` to seed the flags. Right now the database contains `de` and `en` versions. We should display the right one for the user.

Let's set the 20 as an easy to change variable somewhere. I am not sure if 20 is going to be THE number.

briankariuki:
After uploading the photo, the user will be redirected to the interests screen. I've decided to call this `user interests` but a better name for this feature is welcome.

To persist the interests, I've added a table which saves each of the interests by the `flag_id` and `user_id`.

image


lib/animina/accounts/resources/basic_user.ex
```diff
@@ -9,6 +9,7 @@ defmodule Animina.Accounts.BasicUser do
     extensions: [AshAuthentication]
 
   alias Animina.Accounts
+  alias Animina.Traits
```
wintermeyer: We don't need that for the `BasicUser`. Just for the `User`.

lib/animina/accounts/resources/basic_user.ex
```diff
@@ -63,6 +64,11 @@ defmodule Animina.Accounts.BasicUser do
     has_many :photos, Accounts.Photo do
       destination_attribute :user_id
     end
+
+    has_many :interests, Traits.UserInterests do
```
wintermeyer: No need for `BasicUser`.

lib/animina/traits/resources/user_interests.ex
```diff
@@ -0,0 +1,58 @@
...
+
+  code_interface do
+    define_for Animina.Traits
+    define :read
+    define :create
+    define :by_id, get_by: [:id], action: :read
+  end
+
+  preparations do
+    prepare build(load: [:flag])
```
wintermeyer: When we preload the `flag` we can preload the `user` too, or not?

lib/animina/traits/resources/user_interests.ex
```diff
@@ -0,0 +1,58 @@
...
+    table "user_interests"
+    repo Animina.Repo
+
+    references do
+      reference :user, on_delete: :delete
+      reference :flag, on_delete: :delete
+    end
+
+    custom_indexes do
+      index [:user_id]
```
wintermeyer, line 54: We should have a combined index [:user_id, :flag_id] to make sure that it is unique per user.

wintermeyer (changes requested):
Nice work!

Only one little bug: I was able to select more than 20 flags. When that limit it reached I shouldn't be able to select more. They should be disabled to click on.

wintermeyer (approved):

PR #93, opened 2024-02-28
Add and seed a `City` resource.

wintermeyer:


philipp-kempgen:
Late to the party, but are you sure "cities" is a good term here?
Is it meant to also include "towns" and "villages"?
I think I would use "municipalities", which I think comes closer to the meaning of German "Orte".
(I'd use an abbreviation such as "municips" or something.)

Or maybe use "locations", if you want to include not only cities/towns/villages but also locations such as a mountain top.

"Municipality" describes an area. "Location" is more something like a point resp. coordinates.

philipp-kempgen:
Side note: I think that "zip code" is a US-only term.
It gets you into the muddy waters regarding census-designated places, unincorporated areas and US Postal Service designations that many people outside of the US would probably find hard to understand.
Keep in mind that there are even settlements without a zip code.
( https://en.wikipedia.org/wiki/Empire,_Nevada )
I'd use "postal code" as a generic term.

Then again, it's probably not worth changing.

wintermeyer:
> Late to the party, but are you sure "cities" is a good term here?

`cities` is good for now. I see your point but we can fix that in the future.

wintermeyer:
> Then again, it's probably not worth changing.

No, not now. Probably not ever.

philipp-kempgen:
In the end it boils down too what you want to happen when someone enters e.g. "Hennickendorf" (478 inhabitants) as their location. It's part of the "Nuthe-Urstromtal" municipality (340 km<sup>2</sup>, postal code 14947, less than 7000 inhabitants). There aren't any cities. "Ruhlsdorf" would be the political center, but it has even less inhabitants (359).

Someone in Hennickendorf could be displayed as being located "near Ruhlsdorf" ("in Ruhlsdorf" would be wrong) or "near Luckenwalde", but Luckenwalde isn't part of the municipality. The latter still makes sense.

I think the proper way is to display the name of the municipality "Nuthe-Urstromtal". Just keep in mind that there's no city/town/village called "Nuthe-Urstromtal".

Granted you might call those edge cases, as between 70 and 80 % of the German population live in or around cities, but:
- "_around_ cities" doesn't mean "_in_ cities", i.e. different postal code.
- That's still some 20 to 25 million people who don't live in or around cities.
- I have no idea if these people might be less interesting as users, demographically, or even more interesting.

Up to you to decide how much thought to spend here to make later improvements easier.

wintermeyer:
> Up to you to decide how much thought to spend here to make later improvements easier.

Done is better than perfect.

Let's go with the current implementation and cycle back in a couple of months.

PR #140, opened 2024-03-10
Add sign-in page, closes #120, closes #133, closes #139

### LINKED ISSUE #120: PotentialPartnerLive form should load current_user values
Same as #118 

Use the values from `current_user` to fill the form so that we can use it within the registration process and later on as an edit form.

### LINKED ISSUE #133: Auto copy white to green flags during registration
It is very likely that a user wants his/her white flags to become green flags too. So to save him/her work we should automatically select the same flags the first time he/she visits `FlagsLive, :green´.

### LINKED ISSUE #139: Add sign in and sign out functionality
Animina does not currently have a sign in page.

briankariuki:
This PR adds a sign-in page at `/sign-in`. Its pretty basic for now and we can improve it later. It also introduces a sign out route at `/sign-out` which purges any authentication tokens the user has saved in the db. You can suggest where to put the button for signing out.

I've added `on_mount` hooks for liveviews that need `current_user` to be available. You don't have to use `Registration.get_current_user(session)` to check if `current_user` is defined in your liveviews. Simply add the route to the desired `ash_authentication_live_session` in the router and the hooks will set the current_user for you.

I've also added support for populating the `potential_partner_live` page with values the user selected on signup.

image


wintermeyer (approved):
Nice work!

PR #153, opened 2024-03-23
Feature/desktop version of /profile/potential partner , closes #148

### LINKED ISSUE #148: Improve Desktop Version of /profile/potential-partner
Please improve the desktop view for this page and add a dark mode version.

Bildschirmfoto 2024-03-22 um 12 57 47

MICHAELMUNAVU83:
- In this PR , I worked on the desktop version of /profile/potential-partner , I have set a width of 80% on desktop devices as I feel having 100% would be an overkill . I will appreciate your advise on this , below are both screens in dark and light mode .

@wintermeyer  , I decided to go with the 2 column solution , have a look below 
Screenshot 2024-03-24 at 06 19 38

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -4,13 +4,26 @@
 
 /* App is optimized for mobile phones. Limit maximum width on other devices for now. */
 body {
-	max-width: 13.229cm; /* == 500px == 375pt */
-	margin-left: auto;
-	margin-right: auto;
+  margin-left: auto;
+  margin-right: auto;
+}
+
+@media screen and (max-width: 500px) {
+  /* Adjust max-width for smaller screens */
+  body {
+    max-width: 13.229cm;
+  }
+}
+
+@media screen and (min-width: 501px) {
+  /* Adjust width for larger screens */
+  body {
+    width: 80%;
+  }
 }
 
 input:invalid:not(:placeholder-shown) {
-	background-color: hsl( 12, 100%, 95% );
+  background-color: hsl(12, 100%, 95%);
 }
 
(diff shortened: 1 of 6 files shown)


wintermeyer (changes requested):
Not a big fan of the one column solution for the desktop. Can we use the real estate better and do a two or three column solution? I am open to other solutions too.

MICHAELMUNAVU83:
> Not a big fan of the one column solution for the desktop. Can we use the real estate better and do a two or three column solution? I am open to other solutions too.

Okay , would you like me to fill the whole page then , we'd probably have a 3 column solution, or have the full width remain as is and impliment a 2 column solution?

wintermeyer:
> Okay , would you like me to fill the whole page then , we'd probably have a 3 column solution, or have the full width remain as is and impliment a 2 column solution?

Surprise me. Have a look at the other pages and think of a concept which works for all of the registration process. I have no set idea for this problem.

wintermeyer:
@xamiam In case you have time you I'd appreciate your opinion about this problem/design.

xamiam:
Yes, sure! How about making it responsive columns? Just change the first radio element to a dropdown and you'll get 6 elements that resolve nicely in 3, 2 and 1 column(s). 
3 for desktop, 2 for tablet and alike, 1 for mobile.

wintermeyer:
> you'll get 6 elements

The number of elements on this page is not fixed yet. We might add or delete some. Very fluent situation.

xamiam:
Sure. Could still have responsive columns for now. Not a big deal to change later on.

PR #158, opened 2024-03-26
Add user profile page, closes #128, closes #144

### LINKED ISSUE #128: The User Profile LiveView
# Idea

Every `user` has his/her own profile LiveView. Assuming that my username is `wintermeyer` than my profile should be available at https://animina.de/wintermeyer

I am not 100% sure what the content of this profile is going to be. So this is a first attempt to solve the problem. We will adjust it in the future.

## Content

- On top of the page should be the `name`, the age and the `city.name`. `username` would be a nice to have (depending on layout). Maybe the occupation too. All mainly a question of design and screen real estate. 
- After that the profile contains of all the `stories` of the user sorted by `position` and all the white tags.
- I am not sure about the white tags. They have to be listed but I don't know where they make most sense. I was thinking either after the first story, the last story or spliced by categories between stories.

## Misc

At the end of the registration process the profile of the `current_user` should be displayed.

### LINKED ISSUE #144: Deactivate already used headlines in the pull down
Initially I thought it would be a good idea to give users the chance to use the same headline multiple times. I don't think so any more. Please add a validation that this can not happen and deactivate those entries in this pull down. Let them still in it but just grey.

Bildschirmfoto 2024-03-20 um 07 49 05

briankariuki:
We now have a user profile page, accessible from `/username`. 

The user details are displayed in a card with their profile picture, name, city and occupation.

image

The stories are presented in a list with each story presented in a card.

image

Then the user's white flags are shown after their stories.

image

@wintermeyer Feel free to suggest any changes to make this page better.


assets/js/app.js
```diff
@@ -21,8 +21,11 @@ import "phoenix_html";
 import { Socket } from "phoenix";
 import { LiveSocket } from "phoenix_live_view";
 import topbar from "../vendor/topbar";
+import LocalTime from "./date";
```
wintermeyer: What is this used or? I am not a big fan on any JavaScript which is not essentiell.

wintermeyer (changes requested):
Please give me some background about the JavaScript part or try to solve it within Elixir.

assets/js/app.js
briankariuki: The hook just formats a datetime field. I wanted to display a time for when the story was created. But later found out we don't have timestamps for the story resource. I can go ahead and add timestamps or remove this hook entirely.

briankariuki: With phoenix, I haven't found an easier way to get the user's timezone. Most implementations rely on adding the timezone from the browser to the user's session and then using it in elixir land. Either way some javascript is required. A javascript hook for me is much easier as it handles all the formatting and timezone stuff.

wintermeyer: > I can go ahead and add timestamps or remove this hook entirely.

Please do add the timestamps but also remove this hook and don't display the timestamps at all. Maybe in the future we can use them but right now we don't need that information in the GUI.

wintermeyer: > With phoenix, I haven't found an easier way to get the user's timezone. 

Yes, that is a tricky one. But I am very sensitive with adding any JavaScript on the client side. I have had so much trouble with that in the past that I avoid it where possible.

briankariuki: This is now done

PR #183, opened 2024-04-03
Add photo to a dummy account, closes #177

### LINKED ISSUE #177: Add photo to a dummy account
`Mix.Tasks.CreateDummyAccounts` creates dummy accounts for easier development.

Right now a photo is missing. There is `random_photo_url/1` which gives a URL from a portrait but I don't know how to use that to create the avatar photo.

briankariuki:


lib/animina_web/components/layouts/app.html.heex
```diff
@@ -4,13 +4,11 @@
 
 <.desktop_top_text />
 
-<div class="w-[100%] pt-4  flex justify-center">
+<div class="w-[100%] pt-4  flex justify-start items-start">
```
wintermeyer: This should be in a separated PR. It is not relevant for #177

lib/animina_web/live/flags_live.ex
```diff
@@ -34,7 +34,7 @@ defmodule AniminaWeb.FlagsLive do
     socket
     |> assign(page_title: gettext("Select your own flags"))
     |> assign(color: :white)
-    |> assign(navigate_to: "/profile/flags/green")
+    |> assign(navigate_to: "/profile/green-flags")
```
wintermeyer: `/profile/flags/green` is correct. That shouldn't be undone with this PR. Same for the other flag routes in this PR.

lib/animina_web/live/profile_live.ex
```diff
@@ -165,11 +170,11 @@ defmodule AniminaWeb.ProfileLive do
         </div>
 
         <div class="mt-8 space-y-4">
-          <h2 class="font-bold dark:text-white text-xl">My White Flags</h2>
+          <h2 class="font-bold text-xl">My White Flags</h2>
```
wintermeyer: Always use some form of getext for those texts. Otherwise we can not translate it.

wintermeyer (changes requested):

lib/animina_web/components/layouts/app.html.heex
briankariuki: This was from merging main into my branch.

lib/animina_web/live/flags_live.ex
briankariuki: I've reverted this

PR #184, opened 2024-04-03
 Height visualisation , closes #138

### LINKED ISSUE #138: Visualisation of Height
The height of a potential partner is a very important information for most users. I want to visualise this on the profile page to "see" it right away. Have a look at https://max-interactive.com/projects/people/ by @xamiam to get an idea of the concept. The current_user and the profile person stand next to each other in front of a visible 2,00 m and 1,50 m mark. The basic idea:

Bildschirmfoto 2024-03-06 um 14 48 42

It should be placed above the fold on a mobile phone. But it can be small. 

I will send @briankariuki additional files which GitHub doesn't allow me to attach to this issue.

MICHAELMUNAVU83:
- Hello @wintermeyer  ,On the profile page , we show a visualization for the user's height and their preferred potential partner height. The SVGs got tricky so i decided to use CSS , it is now as such .
Screenshot 2024-04-03 at 15 44 19
Screenshot 2024-04-03 at 15 44 43

### DIFF
--- a/lib/animina_web/components/animina_components.ex
+++ b/lib/animina_web/components/animina_components.ex
@@ -179,11 +179,107 @@ defmodule AniminaWeb.AniminaComponents do
     <.flag_card_loading />
   """
 
-  def current_user_stick_figure(assigns) do
+  def height_visualization(assigns) do
     ~H"""
-    <%!-- <%= if @current_user.gender == male  do %>
+    <%= if @current_user.gender == "male"  || @current_user== "female" do %>
+      <div class="flex gap-16 p-4  py-8 justify-start items-end">
+        <div class="flex gap-0">
+          <p class="h-[200px] dark:bg-white bg-black w-[2px] " />
+          <div class="h-[100%] flex  relative">
+            <p class="absolute dark:text-white -top-[20px]  pb-2">200cm</p>
+
+            <p class="md:w-[400px] w-[300px] absolute top-[2px] mb-[180px] h-[1px] dark:bg-white bg-black" />
+            <p class="absolute top-[32px] dark:text-white  pb-3">150cm</p>
+            <p class="md:w-[400px] w-[300px]  absolute top-[52px]  h-[1px] dark:bg-white bg-black">
+            </p>
+
+            <p class="absolute top-[80px] dark:text-white  pb-3">100cm</p>
+            <p class="md:w-[400px] w-[300px]  absolute top-[103px] h-[1px] dark:bg-white bg-black">
+            </p>
+            <p class="absolute top-[132px] dark:text-white  pb-3">50cm</p>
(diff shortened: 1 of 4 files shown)


lib/animina_web/components/animina_components.ex
```diff
@@ -170,4 +170,116 @@ defmodule AniminaWeb.AniminaComponents do
...
+  Flag card loading.
+
+  ## Examples
+
+    <.flag_card_loading />
+  """
+
+  def height_visualization(assigns) do
+    ~H"""
+    <%= if @current_user.gender == "male"  || @current_user== "female" do %>
```
wintermeyer: This does not cover the case when the profile itself is diverse. We want to exclude that case for now too.

lib/animina_web/live/profile_live.ex
```diff
@@ -12,6 +12,12 @@ defmodule AniminaWeb.ProfileLive do
 
   @impl true
   def mount(%{"username" => username}, %{"language" => language} = _session, socket) do
+    average_potential_partner_height =
```
wintermeyer: What do you mean by "average" here?

wintermeyer (changes requested):

lib/animina_web/live/profile_live.ex
MICHAELMUNAVU83: Here , I am getting the average between the maximum height and the minimum height that the user selected

lib/animina_web/components/animina_components.ex
MICHAELMUNAVU83: Just to get it right @wintermeyer  , we should display the height visualization when the profile is only male or female . if not , if the profile is diverse , how should we approach it

lib/animina_web/live/profile_live.ex
wintermeyer: Why do we need that information? The current_user has only access to profiles which fit his settings.

lib/animina_web/components/animina_components.ex
wintermeyer: We only display this visualisation if all participants in it are either male or female.

lib/animina_web/live/profile_live.ex
MICHAELMUNAVU83: I understand now , the height comparison should be between  the profile being viewed and the current user looking at the profile

wintermeyer: > I understand now , the height comparison should be between the profile being viewed and the current user looking at the profile

Yes! It should be a quick way to "feel" the height of the profile compared to myself.

MICHAELMUNAVU83: @wintermeyer  , thanks for this explanation , I have made tha changes for this , the current user on the left and the profile being viewed on the right .

PR #187, opened 2024-04-04
get random potential partners, closes #186, closes #188

### LINKED ISSUE #186: Return 10 random users. Just to get things started.

### LINKED ISSUE #188: Return all the users which can be selected by everything which is not flag and not GEO related (e.g. query for female and height, etc.)

briankariuki:


lib/animina_web/potential-partner.ex
```diff
@@ -0,0 +1,20 @@
+defmodule AniminaWeb.PotentialPartner do
```
wintermeyer: Shouldn't be the filename `potential_partner.ex`and not `potential-partner.ex`? I am not a 100% sure but `_` feels better to me.

lib/animina_web/potential-partner.ex
```diff
@@ -0,0 +1,20 @@
...
+
+  alias Animina.Accounts
+  alias Animina.Accounts.User
+
+  require Ash.Query
+
+  @doc """
+  Gets random users by the given limit.
+  """
+  def get_random_users(limit \\ 10) do
```
wintermeyer: We should be able to call this function with a `current_user` argument to make sure that this user is not part of the result.

lib/animina/accounts/resources/user.ex
```diff
@@ -104,6 +106,12 @@ defmodule Animina.Accounts.User do
 
   actions do
     defaults [:create, :read, :update]
+
+    read :random_users do
```
wintermeyer: Why do we need this action?

wintermeyer (changes requested):

lib/animina/accounts/resources/user.ex
briankariuki: I've moved this to the PotentialPartner module. Its now part of the query and not an action

lib/animina_web/potential-partner.ex
briankariuki: The function now accepts a user as an argument

PR #211, opened 2024-04-06
Feature/image on height block , closes #209

### LINKED ISSUE #209: Just one block per person - Height comparison
I just want one block per person. That block should represent the height of that person and in that block his/her name should be written. If the name is too long, trim it to the first couple of letters.

If possible: Rotate the font by 180 degrees. The first letter should start at the bottom. But only if possible.

Bildschirmfoto 2024-04-06 um 10 27 32

MICHAELMUNAVU83:
- For this PR , I added the profile picture in each block of the height visualisation.
Screenshot 2024-04-06 at 12 06 25

### DIFF
--- a/lib/animina_web/components/animina_components.ex
+++ b/lib/animina_web/components/animina_components.ex
@@ -304,4 +304,3 @@ defmodule AniminaWeb.AniminaComponents do
     """
   end
 end
-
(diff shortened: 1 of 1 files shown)


wintermeyer:
@MICHAELMUNAVU83 `mix credo` can be a PITA. 😉

Bildschirmfoto 2024-04-06 um 11 20 00

PR #247, opened 2024-04-14
Create a pub/sub channel of every logged in user and trigger updates every 2 seconds , closes #235

### LINKED ISSUE #235: Create a pub/sub channel of every logged in user and trigger updates every 2 seconds
When a user logs into the system he/she should subscribe to his/her own channel. We use this channel for updates.

Add a system wide timer which gets triggered every 2 seconds. It should select the credits which got inserted into the database in the last 2 seconds and update the credit/point information for those users who are logged in. Probably wise to add the currently logged in users to that select query.

MICHAELMUNAVU83:
In this PR , We wanted to ensure a user sees their points tally going up after every 2 seconds .
Whenever a user views a profile , points are added to the profile after every 5 seconds.
For the system wide timer that updates every 2 seconds I used a GenServer that is in `lib/animina/genservers/profileview_credits.ex` for this feature.
I have a PubSub topic called "credits" where logged in users subscribe to and publish to once they view profiles.

The first video shows the points of the profile being viewed going up while the second video shows the user viewing that profile.

https://github.com/animina-dating/animina/assets/86654131/cdc10767-5c06-464c-bc80-83411cf4fb19

https://github.com/animina-dating/animina/assets/86654131/1bd9541e-b82a-445d-b175-c185aa527292


lib/animina_web/live/auth_live/auth_form.ex
```diff
@@ -33,6 +39,23 @@ defmodule AniminaWeb.AuthLive.AuthForm do
...
+    current_user_credit_points =
+      if socket.assigns.current_user do
+        ProfileViewCredits.get_updated_credit_for_user(socket, credits)
+      else
+        0
+      end
+
+    {:noreply,
+     socket
+     |> assign(current_user_credit_points: current_user_credit_points)}
```
wintermeyer, line 52: I don't fully understand this. Why don't we use `current_user.credit_points` for this? Why an additional `current_user_credit_points`?

lib/animina_web/live/story_live.ex
```diff
@@ -3,13 +3,19 @@ defmodule AniminaWeb.StoryLive do
...
   alias Animina.Narratives
   alias Animina.Narratives.Headline
   alias Animina.Narratives.Story
   alias AshPhoenix.Form
+  alias Phoenix.PubSub
 
   @impl true
   def mount(%{"id" => story_id}, %{"language" => language} = _session, socket) do
+    if connected?(socket) do
+      PubSub.subscribe(Animina.PubSub, "credits")
```
wintermeyer, line 16: Shouldn't this subscribe to something like `credits` AND a `user_id`? I only want to get an update for the credits for that user and not for all users of the system or do I?

wintermeyer (changes requested):
Please describe the mechanism you implemented a bit more verbose. I don't understand how it works in a multiuser system. Assuming we have a couple of 1,000 concurrent users in the system. Where does the system trigger the every 2 second update? How does it find out who is subscribed and how are those users updated?

MICHAELMUNAVU83:
@wintermeyer  , the mechanism is as follows , we have a genserver that starts when the system is initialized and we have a regular call that is made by this GenServer every 2 seconds ,   here is the documentation I used https://hexdocs.pm/elixir/1.12/GenServer.html#module-receiving-regular-messages
```
def handle_info(:credits_update, state) do
    schedule_credits_update()

    PubSub.broadcast(Animina.PubSub, "credits", {:display_updated_credits, state})

    {:noreply, state}
  end
```

```
defp schedule_credits_update do
    Process.send_after(self(), :credits_update, 2_000)
 end
```

After every 2 seconds we push all the credits that have been updated to the "credits" topic , the updated credits look like

`[%{"points" => 2, "user_id" => 1}, %{"points" => 5, "user_id" => 2}]`

Now on each liveview , after the 2 secons regular call by the Genserver , we have this handle info call

``` 
def handle_info({:display_updated_credits, credits}, socket) do
    current_user_credit_points =
      ProfileViewCredits.get_updated_credit_for_user(socket, credits)

    {:noreply,
     socket
     |> assign(current_user_credit_points: current_user_credit_points)}
end
```

```
def get_updated_credit_for_user(socket, credits) do
    case Enum.find(credits, fn credit ->
           credit["user_id"] == socket.assigns.current_user.id
         end) do
      nil -> socket.assigns.current_user.credit_points
      credit -> credit["points"]
    end
end
```


We filter through the list to find a user_id for the current user and use the points there , if there is no user_id matching that of the current user in the credits pushed through PubSub , that means his/her points remain the same .

wintermeyer:
Why don't we have a cronjob which runs every 2 seconds and fetches the updates in the database. Select users whose credits got updated in the last 2 seconds. Run though that list of users and updates those accounts. Wouldn't that be a better approach which scales and handles outages better? If not: Why?

wintermeyer:
Follow up question: How does your solution handle if a user gets credits from 100 other users during the 2 seconds?

MICHAELMUNAVU83:
A Genserver would be a cron job running after every 2 seconds , the logic would be the same . What would you want us to use for the cron job instead? 
On the second question , if 100 people are viewing a profile that means in the credits array we will have

`[%{"points" => his/her points + 100 "user_id" => 1},]`,
so we would display that , it would work efficiently.

MICHAELMUNAVU83:
@wintermeyer  , down the road we can look at https://github.com/elixir-wallaby/wallaby to test the concurrency here

PR #260, opened 2024-04-15
Add the number of green and red flags above the fold , closes #240

### LINKED ISSUE #240: Add the number of green and red flags above the fold
Right now I'd have to manually count the number of green and red flags of a given profile. Since this is a very interesting information we should calculate and display it for the current_user right at the top of each profile.

Bildschirmfoto 2024-04-13 um 12 02 22

MICHAELMUNAVU83:
- I added the number of green and red flags above the fold  as such

Screenshot 2024-04-15 at 09 41 14

### DIFF
--- a/lib/animina_web/live/profile_live.ex
+++ b/lib/animina_web/live/profile_live.ex
@@ -45,8 +45,22 @@ defmodule AniminaWeb.ProfileLive do
           {current_user_green_flags, current_user_red_flags} =
             fetch_green_and_red_flags(current_user.id, language)
 
+          intersecting_green_flags_count =
+            get_intersecting_flags(
+              fetch_flags(current_user.id, :green, language),
+              fetch_flags(user.id, :white, language)
+            )
+
+          intersecting_red_flags_count =
+            get_intersecting_flags(
+              fetch_flags(current_user.id, :red, language),
+              fetch_flags(user.id, :white, language)
+            )
+
           socket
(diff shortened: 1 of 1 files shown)


lib/animina_web/live/profile_live.ex
```diff
@@ -45,8 +45,22 @@ defmodule AniminaWeb.ProfileLive do
           {current_user_green_flags, current_user_red_flags} =
             fetch_green_and_red_flags(current_user.id, language)
 
+          intersecting_green_flags_count =
```
wintermeyer, line 48: I don't like how we hit the database multiple times for flag related stuff. We fetch :white, :red, :green and again :red and :green. That is just waisting computing power for nothing. Please refactor and rename `fetch_green_and_red_flags/2` to do one database request for all colours of that given user and split it within Elixir and do all the needed calculations.

wintermeyer: One could even argue that this all should be done in the resource (e.g. with `aggregates`).

@MICHAELMUNAVU83 Fire at will or don't we can kick this refactoring can down the road.

wintermeyer: But I don't want too many database hits. Keep it to a minimum.

wintermeyer (approved):
I'll open up new issues for the stuff I'd like to get done differently.

PR #272, opened 2024-04-16
Impossible flags should be grey , closes #270

### LINKED ISSUE #270: Impossible flags should be grey
Right now one has to hover above an impossible to select flag to see a different color. I am not a 100% sure but let's try to display those flags in a light grey. And no different color for hovering.

Bildschirmfoto 2024-04-16 um 08 59 53

MICHAELMUNAVU83:
Flags that have already been selected by the opposite color are now grey.
Screenshot 2024-04-16 at 16 50 09

Screenshot 2024-04-16 at 16 49 15
Screenshot 2024-04-16 at 16 49 00

### DIFF
--- a/lib/animina_web/components/select_flags_component.ex
+++ b/lib/animina_web/components/select_flags_component.ex
@@ -79,24 +79,28 @@ defmodule AniminaWeb.SelectFlagsComponent do
             aria-label="button"
             phx-click={
               if(
-                get_cursor_styling(
+                get_flag_styling(
                   @can_select,
                   @selected_flags,
                   flag.id,
-                  @opposite_color_flags_selected
+                  @opposite_color_flags_selected,
+                  flag,
+                  @color
                 ) == "cursor-pointer",
                 do: "select_flag",
                 else: nil
               )
             }
-            class={"rounded-full flex gap-2 items-center  px-3 py-1.5 text-sm font-semibold leading-6  focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2  #{get_cursor_styling(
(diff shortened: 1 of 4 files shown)


lib/animina_web/components/select_flags_component.ex
```diff
@@ -79,24 +79,28 @@ defmodule AniminaWeb.SelectFlagsComponent do
...
               )
             }
-            class={"rounded-full flex gap-2 items-center  px-3 py-1.5 text-sm font-semibold leading-6  focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2  #{get_cursor_styling(
+            class={"rounded-full flex gap-2 items-center  px-3 py-1.5 text-sm font-semibold leading-6  focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2  #{get_flag_styling(
               @can_select,
               @selected_flags,
               flag.id,
-              @opposite_color_flags_selected
+              @opposite_color_flags_selected,
+              flag,
```
wintermeyer, line 99: Do we need `flag` and `flag.id`? Wouldn't be `flag` enough?

wintermeyer (changes requested):

lib/animina_web/components/select_flags_component.ex
MICHAELMUNAVU83: Here "flag_id" is for the record from the "flag_traits" table while flag.id is from the "user_flags" table

wintermeyer: Puhhh... a bit misleading. But we have bigger fish to fry right now.

PR #284, opened 2024-04-17
Feature/add preapproved communication only , closes #243

### LINKED ISSUE #243: Disable receiving DM (e.g. from males)
Right now anybody can send a message to anybody else. Some platforms restrict this and especially female users appreciate that functionality. We should add a `preapproved_communication_only` boolean to the user resource. The default should be false. Let's add this form field on the second page of the registration process. Call it "Receive messages only from users who I reapproved with a like."

MICHAELMUNAVU83:
- Added preapproved_communication_only to the users resource that defaults to false and can be toggled on the `/profile/potential-partner` page

Screenshot 2024-04-17 at 13 10 42


wintermeyer:
@MICHAELMUNAVU83 Please put the checkbox on the left side of the text.

PR #307, opened 2024-04-20
New Messages Resource , closes #300

### LINKED ISSUE #300: New Messages Resource
We need a `message` resources for the chat functionality. Right now this is a two way system only. One can send a direct message to an other user and vice versa. Here is my idea of the needed attributes. Please feel free to change it.

- ID (Primary Key)
- SenderID (Foreign Key referencing the ID of the Sender from your existing User Table)
- ReceiverID (Foreign Key referencing the ID of the Receiver from your existing User Table)
- Content (max 1.024 characters)
- Timestamps (inserted_at and created_at)
- read_at (timestamp)

Please don't forget indexes on `sender_id` and `receiver_id`.

I'd like to start using policies with the resource. A user can only create a message if the receiver has `preapproved_communication_only` set to `false`.  Please write tests to ensure the working of this.

MICHAELMUNAVU83:
- Hello @wintermeyer  , I added the messages resource together with the required relationships and indexes , on the next PR , I will workon policies


lib/animina/accounts/accounts.ex
```diff
@@ -13,5 +13,6 @@ defmodule Animina.Accounts do
     resource Animina.Accounts.Token
     resource Animina.Accounts.Photo
     resource Animina.Accounts.Reaction
+    resource Animina.Accounts.Message
```
wintermeyer, line 16: I don't know if `Accounts` is the right home for this resource. But we can move that later.

lib/animina/accounts/resources/message.ex
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    belongs_to :receiver, Animina.Accounts.User do
+      api Animina.Accounts
+      attribute_writable? true
+    end
+  end
+
+  actions do
+    defaults [:create, :read, :update, :destroy]
```
wintermeyer: We don't support `update` and `destroy` right now. Later on maybe but now I want it to be as simple as possible.

lib/animina_web/live/auth_live/user_auth.ex
```diff
@@ -114,11 +97,28 @@ defmodule AniminaWeb.LiveUserAuth do
     end
   end
 
-  def is_multiple_of_ten(streak) when is_integer(streak) do
-    rem(streak, 10) == 0
+  defp add_daily_points_for_user(_user, _points, _) do
   end
 
-  defp add_daily_points_for_user(_user, _points, _) do
+  def check_if_user_has_daily_bonus_added_for_the_day(user_id) do
```
wintermeyer, line 103: Please don't mix multiple topics into the same PR. But leave it for now.

wintermeyer (changes requested):

lib/animina/accounts/resources/message.ex
MICHAELMUNAVU83: OKay , let me make this change

lib/animina_web/live/auth_live/user_auth.ex
MICHAELMUNAVU83: Okay

Issue #262, opened 2024-04-15
Don't hit the database too often for the flags on the profile


wintermeyer:
See https://github.com/animina-dating/animina/pull/260/files/d38eeee8717b7c0c803e31ff0afd8a5b00196606

Bottom line: We don't want to hit the database too often for very similar stuff. In the ideal world we use the `User` resource to fetch the data once and everything we need gets calculated and delivered with that request.

PR #343, opened 2024-04-24
Fix top content on chat live  , closes #337

### LINKED ISSUE #337: Same user information
When I open the chat for http://localhost:4000/wintermeyer/messages/Musterfrau I see this on top of that page:

Bildschirmfoto 2024-04-24 um 08 12 34

That doesn't help me as a user. The information from the profile page would give me a bit more background. So please refactor the following part and not only put it in the profile page but also use that as a replacement for the above blue thing for the chat page.

Bildschirmfoto 2024-04-24 um 08 12 16

PS: Please make sure that the heart is clickable and works on all pages it is used.

MICHAELMUNAVU83:
On the chat live , we have information about the receiver and a button to like and unlike

https://github.com/animina-dating/animina/assets/86654131/15746e41-5211-48c4-850a-30c68beb351b

### DIFF
--- a/lib/animina_web/components/chat/chat_components.ex
+++ b/lib/animina_web/components/chat/chat_components.ex
@@ -4,6 +4,7 @@ defmodule AniminaWeb.ChatComponents do
   """
   use Phoenix.Component
   alias Animina.Markdown
+  alias AniminaWeb.ProfileComponents
 
   def send_message_button(assigns) do
     ~H"""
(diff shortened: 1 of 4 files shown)


lib/animina_web/components/chat/chat_components.ex
```diff
@@ -33,10 +34,19 @@ defmodule AniminaWeb.ChatComponents do
...
     <div class="h-[80vh]  w-[100%] ">
       <div class="w-[100%] h-[100%] flex flex-col justify-between">
-        <div class="h-[7%] w-[100%]">
-          <.receiver_profile_box receiver={@receiver} />
+        <div class="h-[10%] w-[100%] z-50">
+          <ProfileComponents.profile_details
+            user={@receiver}
+            current_user={@sender}
+            current_user_has_liked_profile?={@current_user_has_liked_profile?}
+            profile_points={@profile_points}
```
wintermeyer, line 42: This is over bloated. It should be something like:

`<.ProfileComponents.profile_details user={@receiver} current_user={@sender} />`

We digg our own grave with the current approach. The user resource should auto load the flags, the points, etc. No need to have a million arguments for this which could all be part of `user`.

wintermeyer (changes requested):

lib/animina_web/components/chat/chat_components.ex
MICHAELMUNAVU83: Okay , let me take time to refactor the same thing , same for the profile_live , that should also close #262

wintermeyer: Do you want me to merge the PR now or do you want to refactor first?

MICHAELMUNAVU83: I feel the refactor would be a huge task that would take time and focus , best for morning hours , can we leave this open , I do the other small issues to reduce the  issues then focus on this tommorow morning?

wintermeyer: Let's merge it now and open a new one tomorrow. Done is better than perfect.

wintermeyer:
@MICHAELMUNAVU83 Please create a new issue in case you want one for the refactoring.

MICHAELMUNAVU83:
I have added  #262 to  the MVP milestone

PR #348, opened 2024-04-24
Feature/fix accounts by username function to get exact usernames , closes #334

### LINKED ISSUE #334: Redirect /musterfrau to /Musterfrau
Assuming a user created an account `Musterfrau` than her profile is available at http://localhost:4000/Musterfrau and http://localhost:4000/musterfrau and http://localhost:4000/musterfrAU but actually only the first one is correct.

Please do a proper redirect to the right username route in case there is a misspelling.

MICHAELMUNAVU83:
- Hello @wintermeyer  , to fix this , I had to change the data type of the username to a normal string , previously we had it as such https://hexdocs.pm/ash/Ash.CiString.html#value/1 , which is case insensitive .


wintermeyer (changes requested):
I think we don't need to change the string type in the database which allows us an easier search. Why not do something like this? Pseudocode.

```elixir
    socket =
      case Accounts.User.by_username(username) do
        {:ok, user} ->
          socket
          |> assign(:user, user)
          
          if username != user.username do
            redirect(to: "/#{user.username}")
          end
 ```

MICHAELMUNAVU83:
This would still be an option , is there any advantage to using the ci_string data type?

wintermeyer:
> This would still be an option , is there any advantage to using the ci_string data type?

Our users are non technical people who do not understand that `HouSE` is not the same as `house` for a search algorithm. On the other hand we are dealing with humans who play with that to do something like `JamesBond` and who like to see that this way. So we don't want to downcast everything which we would otherwise have to. 

So at the end of the day the `ci_string` is not a bad solution for the given problem. Everybody can "misspell" it and we just fix it afterwards to the way the original human wanted it to be.

MICHAELMUNAVU83:
@wintermeyer  , we now have it as such both for profile and chat lives , it redirects to the actual username , that was genius from you.

https://github.com/animina-dating/animina/assets/86654131/55ccff7a-4266-46df-ba58-7be8a344fb97

PR #351, opened 2024-04-25
Feature/preload on user resource , closes 262

MICHAELMUNAVU83:
Previously on the Profile live , we would fetch a user , their stories  and flags seperately.
This Pr introduces a way to fetch the user only and this comes with their stories and flags preloaded .


lib/animina_web/live/profile_live.ex
```diff
@@ -307,20 +305,16 @@ defmodule AniminaWeb.ProfileLive do
...
     end)
   end
 
   defp fetch_stories_and_flags(user, language) do
-    stories = fetch_stories(user.id)
+    stories = user.stories
 
     flags =
-      fetch_flags(user.id, :white, language)
+      fetch_flags(user, :white, language)
```
wintermeyer: I have the feeling that `fetch` is a misleading wording here. For me this feels more like a `filter`. Same with other instances in this PR.

wintermeyer (changes requested):
This is the way! Great work!

https://youtu.be/uelA7KRLINA?si=XklVkt8gD8iXb0Sk&t=8

lib/animina_web/live/profile_live.ex
MICHAELMUNAVU83: Yeah , Let me change that for better readability , @wintermeyer  , the PR is now ready

wintermeyer (approved):

PR #354, opened 2024-04-26
Refactored user resource to include flags ,  closes #353

### LINKED ISSUE #353: user has_many flags through user_flags
The following should be a has_many through relationship. We don't need the `user_flags` but the `flags` attached to the `User` resource.

```elixir
    has_many :flags, Traits.UserFlags do
      api Traits
    end
```

All the `current_user.user_flags(each).flag` code everywhere is suboptimal. It should be `current_user.flags`. The `user_flags` part should be invisible to most developers.

But please don't invest more than 4 hours into this. If this is a long thing just add that information to the issue and close it.

MICHAELMUNAVU83:
- In this PR , I fixed the user resource to now load the user's selected `flags` , now we can access with `flag.id `and not `flag.flag.id` and so on.
Apologies for the many failed linters , could not understand the error ash was giving me , I followed the recommendations provided and I had to add

```
   resource Animina.Traits.UserFlags
    resource Animina.Traits.Flag
    resource Animina.Traits.FlagTranslation
    resource Animina.Traits.Category
    resource Animina.Traits.CategoryTranslation
```

to the Animina.Accounts api too.


lib/animina_web/live/chat_live.ex
```diff
@@ -74,8 +74,8 @@ defmodule AniminaWeb.ChatLive do
   end
 
   defp get_intersecting_flags(first_flag_array, second_flag_array) do
-    first_flag_array = Enum.map(first_flag_array, fn x -> x.flag.id end)
-    second_flag_array = Enum.map(second_flag_array, fn x -> x.flag.id end)
+    first_flag_array = Enum.map(first_flag_array, fn x -> x.id end)
```
wintermeyer: Not super important, but I think the following might be easier to read (I am not sure which one is faster):

```elixir
Enum.filter(first_flag_array, &(&1 in second_flag_array))
|> Enum.count
```

lib/animina_web/live/chat_live.ex
```diff
@@ -74,8 +74,8 @@ defmodule AniminaWeb.ChatLive do
   end
 
   defp get_intersecting_flags(first_flag_array, second_flag_array) do
```
wintermeyer: If the result is an Integer (count) than `get_intersection_flags/2` is not a good name.

lib/animina_web/live/flags_live.ex
```diff
@@ -53,10 +53,10 @@ defmodule AniminaWeb.FlagsLive do
     )
     |> assign(
       :opposite_color_flags_selected,
-      fetch_flags(socket.assigns.current_user.id, :white)
-      |> Enum.map(fn flag -> flag.flag.id end)
+      fetch_flags(socket.assigns.current_user, :white)
```
wintermeyer: I don't like the name `fetch_flags/2`. For me this feels more like a `filter_flags/2`.

lib/animina_web/live/profile_live.ex
```diff
@@ -283,33 +282,35 @@ defmodule AniminaWeb.ProfileLive do
   defp fetch_green_and_red_flags(user, language) do
```
wintermeyer: I don't like the name of the function compared to the result of the function. Same with variables within the function. As far as I understand the code it is not `green_flags` but `green_flags_ids`.

wintermeyer (changes requested):

PR #379, opened 2024-05-02
Feature/add pagination to messages  , closes #320

### LINKED ISSUE #320: How to make sure that not 500 messages are loaded
This is an edge case but how can we make sure that there is no scenario where we load 500 or more messages from the past in the current chat window?

MICHAELMUNAVU83:
- In this PR , I added pagination to messages read between a sender and receiver which default to `400` , @wintermeyer , kindly advise if this number is too high , to add a customization to this , someone would call 
`Message.messages_for_sender_and_receiver(sender.id, receiver.id, page: [limit: 10], actor: sender)` changing the limit parameter , the video below  shows messages read defaulting  to 5

https://github.com/animina-dating/animina/assets/86654131/a760f6cf-35ed-465d-883f-dfd910888704


MICHAELMUNAVU83:
@wintermeyer  , @briankariuki  , could you guys kindly run the tests from your end and tell me if they are passing , they are on my side so I do not know what the issue may be

briankariuki:
> * In this PR , I added pagination to messages read between a sender and receiver which default to `400` , @wintermeyer , kindly advise if this number is too high , to add a customization to this , someone would call
>   `Message.messages_for_sender_and_receiver(sender.id, receiver.id, page: [limit: 10], actor: sender)` changing the limit parameter , the video below  shows messages read defaulting  to 5
> 
>  Screen.Recording.2024-05-02.at.12.35.13.mov

I've seen

> * In this PR , I added pagination to messages read between a sender and receiver which default to `400` , @wintermeyer , kindly advise if this number is too high , to add a customization to this , someone would call
>   `Message.messages_for_sender_and_receiver(sender.id, receiver.id, page: [limit: 10], actor: sender)` changing the limit parameter , the video below  shows messages read defaulting  to 5
> 
>  Screen.Recording.2024-05-02.at.12.35.13.mov

@MICHAELMUNAVU83 Kindly check this bit. I think it should be == false or == nil. 

```
if System.get_env("DISABLE_ML_FEATURES") == true do
    describe "Tests for Nsfw photos detection using AI" do
      setup do
```

MICHAELMUNAVU83:
@briankariuki  , in this case I only want to run the tests if I have ML Features enabled hence I have `System.get_env("DISABLE_ML_FEATURES") == true ` now when running tests you have `DISABLE_ML_FEATURES = true mix test`

PR #392, opened 2024-05-03
Fixed Duplicate flags  , closes #382

### LINKED ISSUE #382: Duplicated flags
I registered a new account and added some flags. Flags are displayed duplicated in the profile. That is wrong and should not happen.

Bildschirmfoto 2024-05-02 um 15 32 55

MICHAELMUNAVU83:
- In this PR I fixed the error that showed duplicate flags , we now have it as such
Screenshot 2024-05-03 at 09 26 36

### DIFF
--- a/lib/animina_web/live/chat_live.ex
+++ b/lib/animina_web/live/chat_live.ex
@@ -32,7 +32,7 @@ defmodule AniminaWeb.ChatLive do
     receiver = Accounts.User.by_username!(profile)
 
     {:ok, messages_between_sender_and_receiver} =
-      Message.messages_for_sender_and_receiver(sender.id, receiver.id,  actor: sender)
+      Message.messages_for_sender_and_receiver(sender.id, receiver.id, actor: sender)
 
     intersecting_green_flags_count =
       get_intersecting_flags_count(
@@ -161,6 +161,7 @@ defmodule AniminaWeb.ChatLive do
       |> Enum.filter(fn x ->
         find_user_flag_for_a_flag(user.flags_join_assoc, x).color == color
       end)
+      |> Enum.uniq()
 
     Enum.map(user_flags, fn user_flag ->
       %{
--- a/lib/animina_web/live/chat_live.ex
+++ b/lib/animina_web/live/chat_live.ex
@@ -156,18 +156,17 @@ defmodule AniminaWeb.ChatLive do
   end
 
   defp filter_flags(user, color, language) do
-    user_flags =
-      user.flags
-      |> Enum.filter(fn x ->
-        find_user_flag_for_a_flag(user.flags_join_assoc, x).color == color
+    traits =
+      user.traits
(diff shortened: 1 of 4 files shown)


lib/animina_web/live/chat_live.ex
```diff
@@ -161,6 +161,7 @@ defmodule AniminaWeb.ChatLive do
       |> Enum.filter(fn x ->
         find_user_flag_for_a_flag(user.flags_join_assoc, x).color == color
       end)
+      |> Enum.uniq()
```
wintermeyer: Why do we have to run a `uniq/1` here at all? Shouldn't it be unique right away? Where is the duplication coming from?

wintermeyer (changes requested):

lib/animina_web/live/chat_live.ex
wintermeyer: What I mean: Do we have two entries in the database for the same flag (same color)? Because that should not be happening.

MICHAELMUNAVU83: Let me have a closer look at our query and get back

MICHAELMUNAVU83: I have found the bug 
```
defp find_user_flag_for_a_flag(user_flags, flag) do
    Enum.find(user_flags, fn x -> x.flag_id == flag.id end)
  end
```

this function returns duplicates and in this case a user may have the same white and green flags

MICHAELMUNAVU83: @wintermeyer  , we may run into problems with the current implimentation where we preload individual flags , here is the related issue #353 , I suggest we go back to fetching the flags through the "user_flags" as we can filter explicity using the color.

For example , since someone can have the same white and red flag or white and green flag , when we fetch the user_flags associated with a flag_id , we will get 2 or 3 records .
IMO having the  using the previous one where we get a flag by `flag.flag.id` etc would work best

wintermeyer: OK, go back if needed. But please avoid code which read like `flag.flag.name` like we had in the past. That is not easy to understand. 

I am pretty sure we can find a clever way to add the `flag.name` information to the `user_flag` resource. So that we could do something like ´current_user.user_flags` and not `current_user.user_flags[0].flag`.

Fire at will.

wintermeyer: Should I merge this PR or not?

MICHAELMUNAVU83: No , Let me get at fixing it , in a couple of hours I will have found the best solution

MICHAELMUNAVU83: @wintermeyer  , for readability , would changing the access point name be better  , so we have  `user.traits.flag.name`

wintermeyer (approved):

Issue #404, opened 2024-05-05
<ul> and <ol> indention


wintermeyer:
This is related to #384 

<img width="545" alt="Bildschirmfoto 2024-05-05 um 09 33 52" src="https://github.com/animina-dating/animina/assets/69770/95a7705f-834c-4e4d-b453-276e4f8a1c8f">

- `<ol>` should be numbers
- Both `<ol>` and `<ul>` should be indented a bit to the right. Please have a look at the E of the first sentence. The first bullet point should be intended to the right from that position. It used to work in the past. Something was changed.

Please doublecheck this with stories and in the chat.

## Example

If I use this here (in GitHub) it gets indented.

- a
- b
- c

1. a
2. b
3. c

MICHAELMUNAVU83:
@wintermeyer  , adding vanilla css styles work for this problem

PR #407, opened 2024-05-05
Feature/fix ol styling, #404

MICHAELMUNAVU83:
- Added indendation to ol and ul .
- For ol , we have them as numbers
- For ul , we have them as dots.

Screenshot 2024-05-05 at 14 53 00
Screenshot 2024-05-05 at 14 54 01


wintermeyer (approved):
I wonder if there is a cleaner solution. But done is better than perfect.

PR #409, opened 2024-05-07
Feature/mix create dummy enhancements , closes  #383

### LINKED ISSUE #383: mix create_dummy_accounts should use a default password
Please change `mix create_dummy_accounts` so that those accounts have the standard password `test`.

Bonus: 
- If possible make sure that this mix task can not be run in non dev environments.
- Add `animina` to the list of forbidden passwords.

MICHAELMUNAVU83:
- Added `animina` to the bad_passwords.txt
- Added `test` as the default password
- Ensured the mix task `mix create_dummy_accounts 10` can only be run in dev and test environments

### DIFF
--- a/lib/mix/tasks/create_dummy_accounts.ex
+++ b/lib/mix/tasks/create_dummy_accounts.ex
@@ -1,4 +1,4 @@
-if Mix.env() == :dev || Mix.env() == :test do
+if  Enum.member?([:dev, :test], Mix.env()) do
   defmodule Mix.Tasks.CreateDummyAccounts do
     @moduledoc """
     This task creates dummy accounts for development.
@@ -54,7 +54,7 @@ if Mix.env() == :dev || Mix.env() == :test do
             height: height,
             mobile_phone: random_mobile_phone_number(),
             birthday: birthday,
-            hashed_password: "test",
+            hashed_password: Bcrypt.hash_pwd_salt("test"),
             occupation: Person.En.title_descriptor(),
             minimum_partner_height: height - 30,
             maximum_partner_height: height + 15,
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/create_dummy_accounts.ex
```diff
@@ -1,313 +1,315 @@
...
+            username: Faker.Internet.user_name() |> String.slice(0..14),
+            name: Faker.Person.name(),
+            zip_code: random_zip_code(),
+            language: "DE-de",
+            legal_terms_accepted: true,
+            gender: gender,
+            height: height,
+            mobile_phone: random_mobile_phone_number(),
+            birthday: birthday,
+            hashed_password: "test",
```
wintermeyer: How can the value of the `hashed_password` be `test` if I want the password to be test?

lib/mix/tasks/create_dummy_accounts.ex
```diff
@@ -1,313 +1,315 @@
...
-
-      # create about me story
-      story =
-        Story.create!(%{
-          headline_id: about_me_headline,
-          user_id: user.id,
-          content: random_lorem_ipsum(),
-          position: 1
-        })
+if Mix.env() == :dev || Mix.env() == :test do
```
wintermeyer: Nothing wrong here. I just have the feeling that `Enum.member?/2` is easier to read and faster to understand in this case. `Enum.member?([:dev, :test], Mix.env())`

wintermeyer (changes requested):

lib/mix/tasks/create_dummy_accounts.ex
MICHAELMUNAVU83: Thanks for this oversight , I looked at ash_authentication and noticed they use Bcrypt so we now have
 ` Bcrypt.hash_pwd_salt("test")` , I have tested this locally and it works

lib/mix/tasks/create_dummy_accounts.ex
MICHAELMUNAVU83: I have adjusted this code for better readability .

PR #423, opened 2024-05-11
Feature/deduct points on profile visit , closes #417

### LINKED ISSUE #417: Deduct credit points for visits
A user has to pay points to visit a profile. Only public profiles are free to visit. I don't know how much we will charge for that. Let's assume - for now - we charge 20 points for the first visit. Following visits are free of charge.

Please think about this problem and come up with a solution which works for a 3rd party API too.

I guess policies are the best angle of attack for this. We have to make sure that an actor has enough points to read a user resource.

Use the `Credit` resource with a -20 points entry for the bookkeeping.

- [ ] Add an action in the user resource  `by_username_as_an_actor` that we use to check if a user has enough points to view a profile , in this case , if the profile has liked the user and this is the first time , they need at least 10 points , if they have not likes , 20 points
- [ ] For anonymous users they should call `by_username` as there is no actor in this case thus no deduction
- [ ] Add a credit with points of either -10 or -20 for first profile views

MICHAELMUNAVU83:
In the PR Below , I did the following

- [x] - If a user does not have enough credit points , they see an errror message as such "This profile either doesn't exist or you don't have enough points to access it. You need 20 points to access a profile page"
- [x] - I also added a policy to check that if the profile has likes the user , I need a minimum of 10 points to view it
- [x] - I ensured that 20 points are deducted for the first profile visit

### DIFF
--- a/lib/animina/accounts/accounts.ex
+++ b/lib/animina/accounts/accounts.ex
@@ -14,8 +14,8 @@ defmodule Animina.Accounts do
     resource Animina.Accounts.Photo
     resource Animina.Accounts.Reaction
     resource Animina.Accounts.Message
-    resource Animina.Accounts.UserRole
     resource Animina.Accounts.Role
+    resource Animina.Accounts.UserRole
     resource Animina.Traits.UserFlags
     resource Animina.Traits.Flag
     resource Animina.Traits.FlagTranslation
--- a/lib/animina/accounts/resources/credit.ex
+++ b/lib/animina/accounts/resources/credit.ex
@@ -38,12 +38,14 @@ defmodule Animina.Accounts.Credit do
         allow_nil? false
       end
 
-      argument :profile_id, :uuid do
+      argument :donor_id, :uuid do
         allow_nil? false
       end
 
-
-      filter expr(donor_id == ^arg(:donor_id) and user_id == ^arg(:user_id) and subject == "Profile View")
+      filter expr(
+               donor_id == ^arg(:donor_id) and user_id == ^arg(:user_id) and
+                 subject == "Profile View"
+             )
     end
   end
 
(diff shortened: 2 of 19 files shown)


wintermeyer:
The error message should be something like "This profile either doesn't exist or you don't have enough points to access it. You need 20 points to access a profile page."

We need to make sure that there is no way to find out if a specific profile page exists by playing with the system. We are going to have a black list mechanism und that has to be solid. There will be users who don't want specific other users to be able to find out what they are doing (e.g. ex partner, current partner, work colleague).

wintermeyer (approved):

philipp-kempgen:
> We need to make sure that there is no way to find out if a specific profile page exists by playing with the system.

Yes.

> We are going to have a black list mechanism und that has to be solid. There will be users who don't want specific other users to be able to find out what they are doing (e.g. ex partner, current partner, work colleague).

Slightly off-topic here, but since you mentioned it: A blacklist will be full of holes, like swiss cheese. You would have to enter all e-mail addresses (or domains of e-mail addresses) that you want to hide from. You may not know the personal e-mail addresses of your work colleagues, and even then, that's easy to circumvent for someone by creating an additional account using a different address.

You could have an additional means to hide from anyone in specific locations (such as your own location or workplace), by their claimed location or by their IP address. Could be bypassed by someone determined as well.

I think that for now you need to go without a hide list, as it's far from trivial to come up with something that works most of the time and at least represents an obstacle.

wintermeyer:
> A blacklist will be full of holes

Since we are going to use e.g. mobile phone numbers for the black list it will be pretty good and end user friendly. Most people don't want their ex- or current partners to see their profile. For that a simple black list ist more than enough.

philipp-kempgen:
Yeah, validated mobile phone numbers would be pretty good.

PR #442, opened 2024-05-15
Feature/fix id not found in nil bug , closes #432

### LINKED ISSUE #432: connection not available and request was dropped bug
We have some performance or database issue with the current version. When I start to use the application and want to sign in I get the following error:

Bildschirmfoto 2024-05-13 um 21 48 16

MICHAELMUNAVU83:
- Fixed the id not found in nil bug , @wintermeyer ,  it should now be okay , we have
Screenshot 2024-05-15 at 10 46 16
Screenshot 2024-05-15 at 10 56 03


wintermeyer:
@MICHAELMUNAVU83 Don't we have a test to catch this before deployment? If not: Please write one.

MICHAELMUNAVU83:
@wintermeyer  , I tested that an anonymous user sees the profile name , I left out the story , my apologies .
On writing the tests though , I only see the loading components and not the story yet , @briankariuki  and I have discussed this and it seems to be an issue with testing  async streams ,  I will add this as a seperate issue

wintermeyer:
Testing for an HTTP 200 code would be better already. Right now this would result in a 404 in production.

PR #467, opened 2024-05-22
Feature/add 404 to profile and write tests ,  closes #446

### LINKED ISSUE #446: Integration test for user profile
Write an integration test for the following scenarios. Setup: Create a profile for the users `alice` and `bob`. `alice` being a not public profile. `bob` being a public profile.

- Anonymous gets a 200 for `/bob`
- Anonymous gets a 404 for `/alice`
- Bob logs in and gets a 200 for `/bob` and `/alice`
- Bob logs in and logs out and after that gets a 200 for `/bob` and a 404 for `/alice`

There might be some overlapping with #444 I just want to make sure with this test that the profile page itself works and that we don't run into any nil this and nil that issues which result in 404 without being catched by the CI.

MICHAELMUNAVU83:
- Added a custom 404 error html to be displayed , it is as such
Screenshot 2024-05-22 at 05 19 39

@wintermeyer , something to note , this will be seen only in production when we set ` debug_errors: false `in dev.exs.
In development we will see
Screenshot 2024-05-22 at 05 29 19

- I added tests for this to ensure we check for the correct html response


wintermeyer (changes requested):
I am writing this review on my phone. Without having seen the 404 page ever. Just the screenshot. 

The user will be lost on that 404 page. If he is already logged in he should the the normal page design with the normal navigation and not a black page with no links and nothing.

MICHAELMUNAVU83:
Thanks for that , @wintermeyer  , worked on the changes you requested, we now have it as such

Screenshot 2024-05-22 at 08 51 01

Screenshot 2024-05-22 at 08 52 32

PR #468, opened 2024-05-22
Feature/add price list , closes #424

### LINKED ISSUE #424: Price list for points
The more we use points to pay for stuff the more I realise that we need a better mechanism to handle how much something costs or how many points I get for something. Hard coding is too limited.

Example: There will be times with very low traffic and times with very high traffic. Wouldn't it be cool to have a tool to raise prices in peek times and lower prices in slow times?

But I don't want to invest too much time for this right now. I just want to make sure that we don't hard code point prices everywhere.

Please create one module which we can use to hard code the prices for now but which we can change in the future to make it more flexible. We should call that module/function in the rest of the code base when we need the amount of points needed for a specific thing (plus and minus points).

MICHAELMUNAVU83:
- Added a price list module with  the following

- [x] price
- [x] points
- [x] action

We pass in the current user in the `get_price_for_action(action, _current_user)` function , as you suggested @wintermeyer  , different users may have different prices for points


wintermeyer (changes requested):
I meant something different. I want a `get_point_price_for_action(action, current_user || nil)`. 

Right now we have fixed point prices and I want to be able to change that in the future. To do that it would be handy to store all point prices in one module.

`get_point_price_for_action(:view_profile)` would result in 20. Maybe a VIP user gets a different point price `get_point_price_for_action(:view_profile, vip_user)` could be 10.

MICHAELMUNAVU83:
Hello @wintermeyer  , we now have

```
 def get_price_for_action(action, current_user \\ nil) do
    # if the current user is there do something different
    case Enum.find(@points_price_list, fn x -> x.action == action end) do
      nil -> {:error, "Action not found"}
      action -> {:ok, action.price}
    end
  end
``` 
This format will allow us flexibility once we decide which users get which points , I would rather we get the result back as  a tuple for error handling to avoid scenarios like x not found in nil


Now in the PointsPriceList module , we can update this array as we wish down the line

```
@points_price_list [
    %{
      points: 100,
      price: 10,
      action: :view_profile
    },
    %{
      points: 200,
      price: 20,
      action: :send_message
    },
  ]
```

wintermeyer:
But `@points_price_list` list doesn't make sense (we don't need a `price`):

> ```
> @points_price_list [
>     %{
>       points: 100,
>       price: 10,
>       action: :view_profile
>     },
>     %{
>       points: 200,
>       price: 20,
>       action: :send_message
>     },
>   ]
> ```

This would make sense:

```
@points_price_list [
    %{
      points: 100,
      action: :view_profile
    },
    %{
      points: 200,
      action: :send_message
    },
  ]
```

I guess I used the wrong wording. With `price` I did mean amount of needed `points`.

MICHAELMUNAVU83:
Okay , I am curious @wintermeyer  will the price of points be dynamic . Will for example 10 points always cost a euro

wintermeyer:
> Okay , I am curious @wintermeyer will the price of points be dynamic . Will for example 10 points always cost a euro

We are talking about totally different things. I am not searching for a "how can I buy points and how much will that cost per point" solution but for a "how many points will a transaction cost" function. I just want to have a central module which we can adjust in the future to create an eco system where the needed points for a transaction can fluctuate. On a slow day things could become cheaper. Maybe on a Monday a view will only cost 18 points but on a Tuesday it costs 25 points (just an example). I just want to establish a central point to give us a chance to implement such a system in the future.

MICHAELMUNAVU83:
Okay , I have changed the module name to  `Animina.ActionPointsList` and 
```
 @action_points_list [
    %{
      points: 100,
      action: :view_profile
    },
    %{
      points: 200,
      action: :send_message
    },
  ]
``` 
and `get_points_for_action` which I now beleive fits our test case better

MICHAELMUNAVU83:
@wintermeyer  , kindly have a look now and advise

wintermeyer (approved):

PR #515, opened 2024-06-01
Upgrade Erlang to 27.0 and Elixir to 1.16.3

wintermeyer:


wintermeyer:
@briankariuki @MICHAELMUNAVU83 I just upgraded Erlang and Elixir. Please do an `asdf install`

PR #615, opened 2024-06-07
Bug/check photos, closes #578

### LINKED ISSUE #578: Check the photos right away ⏱️
This is a replacement or follow up of #508.

The current system with Oban queuing doesn't scale well. It takes too long to process the photos. A good example is to run `mix seed_demo_system` and see how long it takes to process all photos.

# New approach

Do the photo processing right away after uploading the image. It can be asynchros (the user doesn't have to wait for the result) but it should be triggered right away. Since normal users don't upload a million photos at once this should be a good way to spread the workload. But please use the state machine to give us a chance to restart when things go haywire (e.g. restart of the server).

Get rid of everything Oban for now. We will cross that bridge when we need to.

briankariuki:
This PR replaces oban for photo processing with a genserver that buffers photos in a queue and dispatches them to consumers that handle the processing. It also starts a schedule which checks unprocessed photos and adds them to the queue.


lib/animina/genservers/photo/photo.ex
```diff
@@ -0,0 +1,97 @@
...
+
+  @impl true
+  def handle_info({:fetch_unprocessed_photos, _from}, {queue, pending_demand}) do
+    schedule_fetch_unprocessed_photos()
+
+    bulk_result =
+      Photo
+      |> Ash.Query.for_read(:read)
+      |> Ash.Query.filter(state == ^:pending_review)
+      |> Accounts.read!(authorize?: false, page: [limit: 6])
```
wintermeyer, line 51: We should add an order by `created_at` here. The database otherwise will give any random 6 photos. But it should give us the 6 oldest photos.

wintermeyer (changes requested):

PR #666, opened 2024-06-16
Under Investigation Profile Actions , closes #665

### LINKED ISSUE #665: Under Investigation Profile Actions
Tasks are the following

- For a profile under investigation , 

- [x] They cannot log in
- [x] People cannot view their profile
- [x] Admins Can view their profile
- [x] They are immediately logged out
- [x] We do not use them in the potential partners query

MICHAELMUNAVU83:
In this PR , I did the following and wrote unit and integration tests for them.
For users who are under investigation..
  

- [x] They cannot log in
- [x]   People cannot view their profile
- [x]   Admins Can view their profile
- [x]  They are immediately logged out
- [x]  We do not use them in the potential partners query

@wintermeyer  , If a user who is under investigation tries to sign in , and they add a correct password , we show an error , `Account is under investigation` .
Screenshot 2024-06-16 at 08 57 29

When a user state changes to :under_investigation , they are automatically logged out

https://github.com/animina-dating/animina/assets/86654131/40faacf1-504e-4ada-ae98-7ba1bd94f2d3

### DIFF
--- a/lib/animina/checks/read_profile_check.ex
+++ b/lib/animina/checks/read_profile_check.ex
@@ -12,26 +12,18 @@ defmodule Animina.Checks.ReadProfileCheck do
   end
 
   def match?(actor, params, _opts) do
-    IO.inspect params
-    IO.inspect check_if_user_can_view_profile(actor, params, params.query.arguments) , label: "Mamamia"
+    check_if_user_can_view_profile(actor, params, params.query.arguments)
   end
 
-  defp check_if_user_can_view_profile(_actor, _params, %{}) do
-    IO.inspect("Here s")
-    true
-  end
-
-  defp check_if_user_can_view_profile(actor, params, _) do
+  defp check_if_user_can_view_profile(actor, params, %{username: _username}) do
     case User.by_username(params.query.arguments.username) do
       {:ok, profile} ->
         if actor.username == profile.username || profile.is_private == false do
           user_can_view_profile(
             admin_user?(actor),
             profile.state
-
           )
         else
-          IO.puts("MAmma")
           user_can_view_profile(
             admin_user?(actor),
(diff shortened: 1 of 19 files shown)


lib/animina/checks/read_profile_check.ex
```diff
@@ -37,19 +38,47 @@ defmodule Animina.Checks.ReadProfileCheck do
...
-  defp user_can_view_profile(_, _, _) do
+  defp user_can_view_profile(true, :under_investigation_, _, _) do
+    true
+  end
+
+  defp user_can_view_profile(false, :under_investigation, _, _, _) do
+    false
+  end
+
+  defp user_can_view_profile(false, _, false, true, points) when points >= 10 do
```
wintermeyer: The needed points shouldn't be hard coded here. We have this point priceless module which should be used to define how many points are needed for what to do. That is our central point of defining the needed points for anything. I want that in one place. Otherwise we have it scattered around all over the place. Hard to manage.

wintermeyer (changes requested):
Great work!

Is it much work to trigger a flash message "Your account is currently under investigation.  Please try again to login in 24 hours."

Because right now the user would be automatically logged out but doesn't know why.

lib/animina/checks/read_profile_check.ex
MICHAELMUNAVU83: This has been changed, I now do it from the Actions Price List Module we had prior

MICHAELMUNAVU83:
@wintermeyer , the flash now appears as such
Screenshot 2024-06-16 at 15 48 38

MICHAELMUNAVU83:
@wintermeyer  , I worked on the changes you requested , The PR is now ready

PR #668, opened 2024-06-17
Banned user actions , closes #667

### LINKED ISSUE #667: Banned State Actions
Tasks are the following

- For a profile that is banned , 

- [x] They cannot log in
- [x] People cannot view their profile
- [x] Admins Can view their profile
- [x] They are immediately logged out
- [x] We do not use them in the potential partners query

MICHAELMUNAVU83:
Tasks are the following

- For a profile that is banned , 

- [x] They cannot log in
- [x] People cannot view their profile
- [x] Admins Can view their profile
- [x] They are immediately logged out
- [x] We do not use them in the potential partners query

@wintermeyer  , they get the flash below once they are auto logged out

Screenshot 2024-06-17 at 06 32 21

And the following if they try to login with correct details

Screenshot 2024-06-17 at 11 18 37

### DIFF
--- a/lib/animina/accounts/resources/basic_user.ex
+++ b/lib/animina/accounts/resources/basic_user.ex
@@ -185,6 +185,8 @@ defmodule Animina.Accounts.BasicUser do
     define :create
     define :by_id, get_by: [:id], action: :read
     define :custom_sign_in, get?: true
+    define :investigate
+    define :ban
   end
 
   aggregates do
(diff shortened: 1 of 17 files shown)


wintermeyer (changes requested):
> @wintermeyer , they get the flash below once they are auto logged out

The green "success" flash messages is probably the wrong kind of flash message. It should be a yellow or red error message.

MICHAELMUNAVU83:
I have fixed this @wintermeyer

PR #721, opened 2024-07-04
Nudge the user to understand what is saved in bookmarks , closes #613

### LINKED ISSUE #613: Nudge the user to understand what is saved in bookmarks
When a profile is displayed we could try to nudge the @current_user to understand that he/she will find that profile within the bookmarks again. I tried to draw a little ASCII art with an arrow but failed:

Bildschirmfoto 2024-06-06 um 14 23 34

Please think of something in that way but nicer.

MICHAELMUNAVU83:
- The bookmark tab is highlighted after 3.5 seconds , is active for 3.5 seconds then is deactivated

https://github.com/animina-dating/animina/assets/86654131/078faa32-0cca-4174-bc80-05f7536a7042

### DIFF
--- a/lib/animina_web/components/profile/stories_components.ex
+++ b/lib/animina_web/components/profile/stories_components.ex
@@ -37,7 +37,7 @@ defmodule AniminaWeb.StoriesComponents do
   attr :story, :any, required: true
   attr :photo, :any, required: false
   attr :dom_id, :any, required: false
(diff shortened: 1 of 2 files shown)


wintermeyer:
I am not a big fan of this solution. But I didn't specify what I wanted. Let me close this for now and circle back to it in the future. We have bigger fish to fry right now.

PR #725, opened 2024-07-07
Custom Function to make someone and admin , closes #723

### LINKED ISSUE #723: An admin view where they can make other users admins
@wintermeyer  would this be needed or would you be essentially the only admin?

MICHAELMUNAVU83:
-I added a custom function called `make_admin` that makes a user an admin.
- We use `UserRole.make_admin(%{user_id: user.id})` to do so.


wintermeyer (changes requested):
Please also add a remove admin rights function.

MICHAELMUNAVU83:
I have added this .
Also realized it is better to have this in the user_resource for readability , so now we have
`User.make_admin(%{user_id: user.id})` and `User.remove_admin(%{user_id: user.id})`

PR #764, opened 2024-07-25
Added Change That Creates Optimized Photos After A Photo is created , closes  #656

### LINKED ISSUE #656: We have to rescale the image size and decrease file size
AFAIK we use the uploaded images right now:

Bildschirmfoto 2024-06-11 um 13 48 16

But that is a no no. We can not willy nilly force our users to download > 4MB for a single photo. Ideally we have different resolutions for different things. Like one size for an avatar and one for a normal photo. No image should be bigger than 800Kb.

MICHAELMUNAVU83:
- I added a #TODO  @wintermeyer  , kindly check them out and adjust where necessary.
- If imagemagick is not installed , we copy the image url.


lib/animina/accounts/resources/photo.ex
```diff
@@ -175,4 +195,75 @@ defmodule Animina.Accounts.Photo do
...
+        image_url: record.filename,
+        type: type,
+        user_id: record.user_id,
+        photo_id: record.id
+      })
+    end
+  end
+
+  # TODO: Confirm if the dimensions are  okay for each type
+  defp resize_images_with_imagemagick(record) do
```
wintermeyer: Does it make sense to give a width and a height? We don't know if these are squares. Probably they are not. Just set the width and leave the height to be figured out automatically.

lib/animina/accounts/resources/photo.ex
```diff
@@ -175,4 +195,75 @@ defmodule Animina.Accounts.Photo do
...
+    for type <- [
+          %{
+            width: 100,
+            height: 100,
+            type: :thumbnail
+          },
+          %{
+            width: 300,
```
wintermeyer: Please set the width to 600.

lib/animina/accounts/resources/photo.ex
```diff
@@ -175,4 +195,75 @@ defmodule Animina.Accounts.Photo do
...
+            height: 300,
+            type: :normal
+          },
+          %{
+            width: 800,
```
wintermeyer: Please set the width to 1000.

lib/animina/accounts/resources/photo.ex
```diff
@@ -175,4 +195,75 @@ defmodule Animina.Accounts.Photo do
...
+      OptimizedPhoto.create(%{
+        image_url: record.filename,
+        type: type,
+        user_id: record.user_id,
+        photo_id: record.id
+      })
+    end
+  end
+
+  # TODO: Confirm if the dimensions are  okay for each type
```
wintermeyer: Remove the TODO.

lib/animina/accounts/resources/photo.ex
wintermeyer: Please add that the result will be encoded as a `webp`. We can save a bit more bandwidth with that.

lib/animina/accounts/resources/photo.ex
```diff
@@ -175,4 +195,75 @@ defmodule Animina.Accounts.Photo do
...
+        photo_id: record.id
+      })
+    end
+  end
+
+  # TODO: Confirm if this function is working as expected
+  defp resize_image(image_path, width, height) do
+    image =
+      Mogrify.open(image_path)
+      |> Mogrify.resize("#{width}x#{height}")
```
wintermeyer: Remove the height code.

lib/animina/accounts/resources/photo.ex
```diff
@@ -175,4 +195,75 @@ defmodule Animina.Accounts.Photo do
...
+            type: :normal
+          },
+          %{
+            width: 800,
+            height: 800,
+            type: :big
+          }
+        ] do
+      OptimizedPhoto.create(%{
+        image_url: resize_image(record.filename, type.width, type.height),
```
wintermeyer: Do we need the height? If yes you have to either calculate it or use a tool to figure it out.

lib/animina/accounts/resources/photo.ex
```diff
@@ -175,4 +195,75 @@ defmodule Animina.Accounts.Photo do
...
+    case check_if_image_magick_is_installed() do
+      {:ok, _} ->
+        resize_images_with_imagemagick(record)
+
+      {:error, _} ->
+        copy_images_directly(record)
+    end
+  end
+
+  defp copy_images_directly(record) do
```
wintermeyer: Don't just copy the filename. Copy the file and create a new one. Do it as if you were using imagemagick. The result will be a new file which needs to be stored.

lib/animina/accounts/resources/photo.ex
```diff
@@ -175,4 +195,75 @@ defmodule Animina.Accounts.Photo do
       reference :user, on_delete: :delete
     end
   end
+
+  def create_optimized_photos(record) do
+    case check_if_image_magick_is_installed() do
+      {:ok, _} ->
+        resize_images_with_imagemagick(record)
```
wintermeyer: Please change the function name to `resize_image`. Singular and without the imagemagick part. We might change imagemagick to something else in the future but would keep the function name.

wintermeyer (changes requested):

lib/animina/accounts/resources/photo.ex
```diff
@@ -175,4 +195,77 @@ defmodule Animina.Accounts.Photo do
...
+        image_url: resize_image(record.filename, type.width, type.height),
+        type: type,
+        user_id: record.user_id,
+        photo_id: record.id
+      })
+    end
+  end
+
+  #TODO: Confirm if this function is working as expected
+  defp resize_image(image_path, width, height) do
```
wintermeyer: I don't understand how the new images get new filenames. Do they? Otherwise they would all be the same.

wintermeyer (changes requested):

lib/animina/accounts/resources/photo.ex
MICHAELMUNAVU83: We can use this instead 
`resize_to_fill(image, params)`
Resize the image to fit within the specified dimensions while retaining the aspect ratio of the original image.
If necessary, crop the image in the larger dimension.


SInce you have imagemagick , can you confirm if we can do 
```
 |> Mogrify.resize("#{width}")
```

lib/animina/accounts/resources/photo.ex
MICHAELMUNAVU83: - For image magick , I think it has 2 options , we can either save it in place of the input , or in a different place , My idea was to save it in folder called `optimized` inside `uploads` .
Kindly confirm for me , if you run 

```
 defp resize_image(image_path, width) do
    image =
      Mogrify.open(image_path)
      |> Mogrify.resize("#{width})
      |> Mogrify.save(path: "/uploads/optimized/#{image_path}")

    image.path
  end
```

Seperately and share what it outputs

wintermeyer: The `optimized` path makes sense. But the image files itself have to have different names. Otherwise all three will overwrite each other (because they have the same filename).

lib/animina/accounts/resources/photo.ex
wintermeyer: Assume that it works. I'll test it once everything is put together. In case there is a bug in the Mogrify part I'll fix it.

MICHAELMUNAVU83: Done

lib/animina/accounts/resources/photo.ex
MICHAELMUNAVU83: Done , Thanks

lib/animina/accounts/resources/photo.ex
MICHAELMUNAVU83: That is clear , done

MICHAELMUNAVU83:
@wintermeyer  ,I made the following changes

- We first check if there are the following folders "/uploads/optimized/big" , "/uploads/optimized/normal" and "/uploads/optimized/thumbnail" , if not , we create them.
- If Image magick is not installed , we copy the file to these 3 folders so we would now have  "/uploads/optimized/big/image1 " ,"/uploads/optimized/normal/image2" and "/uploads/optimized/thumbnail/image3" .
- If Image magick is installed , we do the resize and convert to webp , kindly confirm if this works well

wintermeyer (changes requested):
What happens if somebody uploads two different images with the same filename? And what happens if two different users upload images with the same filename? Does the current implementation take care of this or will those images override each other?

MICHAELMUNAVU83:
```
 id: "14fc1d6f-d3af-4c12-9f92-ae388d9e0f35",
  filename: "a175e204-89ef-4260-86d9-7f1211b17a3c.png",
  original_filename: "Featured By (site).png",
  mime: "image/png",
  size: 48937,
  ext: "png",
  dimensions: nil,
  error: nil,
  error_state: nil,
  state: :pending_review,
  created_at: ~U[2024-07-27 06:11:13.588206Z],
  updated_at: ~U[2024-07-27 06:11:13.588206Z],
  user_id: "253cb9ca-1f8b-47ea-8c96-b975dc15c538",
  story_id: "f78fedaa-3c29-4bc4-bae8-71ad90b9edb2",
  aggregates: %{},
  calculations: %{},
```


Seems the filename is a UUID and the original_filename is the name of the actual file so the filename should be unique.

PR #804, opened 2024-07-30
Removed could Interest on mobile , closes #801

### LINKED ISSUE #801: Could interesting you on mobile is right in your face
Have a look at this screenshot for http://localhost:4000 on a mobile devise:

Bildschirmfoto 2024-07-30 um 10 40 03

This doesn't make sense. It should be in the top menu or not there at all.

MICHAELMUNAVU83:
- I removed the could Interest you on Mobile , A user will see this if they have not logged in
Screenshot 2024-07-30 at 12 17 11


wintermeyer:
Never mind. This was related to me not having done the `cd assets && npm install`. Therefor I got this view which didn't make much sense. All is good.

PR #912, opened 2024-08-16
Add button for generated messages to be copied into the input field  , closes #899

### LINKED ISSUE #899: Add button for generated messages to be copied into the input field
It is probably better to give the user an easier way to move the message he wants into the chat field than via cut and paste. Cut and paste works good enough on a desktop but is kind of a hazzle for mobile devises.

So let's add a small button at the end of each message which says "Use this one" and which copies the message into the input field but without already sending it. The user should be able to customise the message. After pressing the button all messages should disappear.

Bildschirmfoto 2024-08-15 um 18 42 19

MICHAELMUNAVU83:
https://github.com/user-attachments/assets/cafd1c23-7a81-4cdc-a0cc-45e1cc81ef47

### DIFF
--- a/lib/animina_web/live/chat_live.ex
+++ b/lib/animina_web/live/chat_live.ex
@@ -183,33 +183,27 @@ defmodule AniminaWeb.ChatLive do
   end
 
   def handle_event("generate_message_with_ai", _params, socket) do
-    # case ChatCompletion.request_message(socket.assigns.sender, socket.assigns.receiver) do
-    #   {:ok, message} ->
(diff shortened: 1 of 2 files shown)


wintermeyer:
It is not clear for which text the "Use this" button is. Of course I know but a normal user might not. Either put it right after the last word of the sentence (as a smaller button) or put it right next to the text.

Bildschirmfoto 2024-08-16 um 11 43 30

MICHAELMUNAVU83:
How about this now?
Screenshot 2024-08-16 at 13 14 26


Screenshot 2024-08-16 at 13 14 04

PR #923, opened 2024-08-20
Delete my account functionality , closes #922

### LINKED ISSUE #922: Delete my account functionality
The user should be able to delete his/her account without any fuzz. We ask him/her one time with a modal if he/she is sure that he/she wants to delete the account and than we do it.

MICHAELMUNAVU83:
- I added this under the profile visibility tab  , I also added unit and integration tests for this feature

https://github.com/user-attachments/assets/c7ff7e49-6db0-4eec-80f7-65802e6ea5f7


wintermeyer (changes requested):
The current modal is not enough. It is too easy to delete an account by accident. hmmm... how can we solve this? I like the position in the menu. But what about linking that link not to the modal but to an extra page which says "You can delete your account here but you have to wait 10 seconds to make sure that this is not an accident." Than a 10 second timer which counts visibly down to 0 and than activates a delete button. That one will trigger the modal and THAN we delete.

It would be create if it was impossible to get to that page without clicking on the link.

@MICHAELMUNAVU83 this is my bad! Sorry for the extra work.

MICHAELMUNAVU83:
That Makes sense , let  me get to it

MICHAELMUNAVU83:
We now have this @wintermeyer  , kindly advise if it suffices

https://github.com/user-attachments/assets/b40d397f-9395-4d19-8963-ae27812bb83f

wintermeyer (approved):

PR #924, opened 2024-08-20
updates on top navigation, closes #921

### LINKED ISSUE #921: Could Interest You: Clickable area is much smaller than the visual button
## Expected
This is the visible area that will change the cursor and change its background colour on hover.

## Actual
This is the clickable area which will navigate to the profile.

## Other issues
* For accessibility `<button>` shouldn't be used. It's all a link `<a>` (because it navigates). It may still look like a button if that's what the designer prefers.

Kimutai01:
https://github.com/user-attachments/assets/427eca3b-5c7c-404e-a8cd-156c5ac357e7


wintermeyer (approved):

wintermeyer:
Good work!

PR #930, opened 2024-08-20
Chat completions, close #906

### LINKED ISSUE #906: Chat: Generating the messages like ChatGPT
Right now we wait till Llama3.1 is finished with the answer until we display the messages. And this takes a lot of time. So much that some people will move on because they think the system is broken.

Is there a chance that we display the messages as they are generated? Word by word?

Please move it to a later milestone in case this is a lot of work.

Kimutai01:


Kimutai01:
https://github.com/user-attachments/assets/09205463-e4c0-4fcc-9566-c3d8be53a6d0

This is the current implementation
Though it's really slow on my laptop

wintermeyer (changes requested):
I really like it!

Just one tiny improvement: Please don’t display the button for each message before that message is finished.

Kimutai01:
https://github.com/user-attachments/assets/ea7bdf05-64ed-40be-9770-55698777ac65

This is how it works,
I had to edit the video because my laptop was hanging a lot but we can test it better on the server

lib/animina_web/live/chat_live.ex
```diff
@@ -416,6 +402,68 @@ defmodule AniminaWeb.ChatLive do
...
+      case ChatCompletion.request_message(socket.assigns.sender, socket.assigns.receiver) do
+        {:ok, task} ->
+          socket
+          |> assign(:current_request, task)
+          |> assign(:generated_message, "")
+          |> assign(:show_use_ai_button, true)
+
+        {:error, _} ->
+          socket
+          |> put_flash(:error, "Could not generate message with AI, Kindly Try Again")
```
wintermeyer, line 455: Always use `gettext/1` for any kind of message.

wintermeyer (approved):
I'll deploy the code on production but please fix the gettext/1 issue.

lib/animina_web/live/chat_live.ex
Kimutai01: Thanks I'll always use it

PR #942, opened 2024-08-22
Deactivate profile if no stories  , close #933

### LINKED ISSUE #933: A profile without a single story shouldn't be active
I saw a profile which doesn't have a story. That shouldn't be possible. At least that profile shouldn't be listed and available for others to see in the system.

MICHAELMUNAVU83:
- If a user has no story , they can visit their profile , but they see
Screenshot 2024-08-21 at 08 50 10

- If Other Users Try to visit that page , they see the following 404 message.
Screenshot 2024-08-21 at 08 59 16


wintermeyer (changes requested):
This kind of logic doesn't grow well into the future. Two problems come to my mind right away:
- What about the already existing profiles? We didn't tell them than they HAVE to create a "About me" story. So they are under the impression that their account is fine. We can not just deactivate those accounts.
- What if in the future we change those rules? May we'll want two stories minimum. Same problem as above.

Please add an `registration_completed_at` timestamp attribute to the `User` resource which is set to nil by default and gets set to current time once the registration process is completed which would include one story for now. And add a migration which sets all the already existing profiles to today 09:00 am so that we can filter them in case we need to in the future.

Change the rules to filter for that attribute not being nil.

MICHAELMUNAVU83:
That makes alot of sense , let me work on these changes.


@wintermeyer  ,If a user goes on to delete all stories , should we then set `registration_completed_at` to nil?

MICHAELMUNAVU83:
@wintermeyer  , I have made the changes , we use the story resource after actions to update the `registration_completed_at` value .

lib/animina/narratives/resources/story.ex
```diff
@@ -172,4 +183,21 @@ defmodule Animina.Narratives.Story do
...
+  defp update_user_registration_completed_at(user_id) do
+    user = Accounts.User.by_id!(user_id)
+
+    case by_user_id(user_id) do
+      {:ok, stories} ->
+        if Enum.count(stories) >= @number_of_stories_required_for_complete_registration &&
+             user.registration_completed_at == nil do
+          Accounts.User.update(user, %{registration_completed_at: DateTime.utc_now()})
+        else
+          Accounts.User.update(user, %{registration_completed_at: nil})
```
wintermeyer: Why would one ever update `registration_completed_at` to `nil`? That attribute is `nil` by default and can only be updated once with a timestamp. After that it is set in stone.

lib/animina/narratives/resources/story.ex
```diff
@@ -164,6 +167,14 @@ defmodule Animina.Narratives.Story do
     identity :unique_position, [:position, :user_id]
   end
 
+  changes do
+    change after_action(fn changeset, record, _ ->
+             update_user_registration_completed_at(record.user_id)
+             {:ok, record}
+           end),
+           on: [:create, :destroy]
```
wintermeyer, line 173: Why on `:destroy`?

lib/animina/narratives/resources/story.ex
```diff
@@ -1,6 +1,9 @@
 defmodule Animina.Narratives.Story do
+  alias Animina.Accounts
   alias Animina.Validations
 
+  @number_of_stories_required_for_complete_registration 1
```
wintermeyer: I am not a big fan of setting this value here. What when we need that number somewhere else too? e.g. in a template where we give the user some sort of feedback. It would be better to set it a default in `config.exs` with the possibility to override it for `prod` or `dev` environment.

lib/animina/accounts/resources/user.ex
```diff
@@ -286,7 +288,8 @@ defmodule Animina.Accounts.User do
         :last_registration_page_visited,
         :streak,
         :confirmed_at,
-        :is_in_waitlist
+        :is_in_waitlist,
+        :registration_completed_at
```
wintermeyer, line 294: Do we have a validation that this value can't be set to a value which is too fare in the past (more than a minute) and too fare in the future (more than a minute)?

priv/repo/migrations/20240822072422_migrate_resources5.exs
```diff
@@ -0,0 +1,26 @@
...
+  use Ecto.Migration
+
+  def up do
+    alter table(:users) do
+      add :registration_completed_at, :utc_datetime_usec
+    end
+
+    execute """
+    UPDATE users
+    SET registration_completed_at = date_trunc('day', now()) + interval '9 hours';
```
wintermeyer: Please set it to a specific timestamp and not some sort of `now()`. Like today at 09:00 am. The 22th of August 2024 at 09:00 am. Otherwise we will not be able to find those entries reliable in the future.

wintermeyer (changes requested):
Now we work with a running production system. We have to think twice. Please go through my review and double check for stuff that I might have forgotten.

lib/animina/narratives/resources/story.ex
MICHAELMUNAVU83: I had previously asked.
If I go an delete all my stories , my account becomes inactive doesn't it?

lib/animina/narratives/resources/story.ex
MICHAELMUNAVU83: @wintermeyer , kindly provide clarity on what should happen if I delete all my stories

lib/animina/narratives/resources/story.ex
wintermeyer: Can I delete all stories? AFAIK I can not delete all but one story or can I?

wintermeyer: I maybe wrong here but AFAIK we implemented a rule which says that there always has to be an `About me` story.

MICHAELMUNAVU83: The rule we had was...you cannot delete the 'About me' story...if it the last one remaining, so if the last one remaining is not a an 'About me" story , then we can delete it and have no stories at all...should we change this rule?

wintermeyer: Of course. We always want to have at least one `About me` story. That is the minimum.

MICHAELMUNAVU83: Someone can have one story per headline . So they have one `About me` story . This new rule will mean , the user cannot delete all their stories , they have to remain with at least one?

wintermeyer: Yes, the `About me` story is the corner stone. A profile has to have at least this story.

priv/repo/migrations/20240822072422_migrate_resources5.exs
MICHAELMUNAVU83: Okay  , I will set it up like this

lib/animina/narratives/resources/story.ex
MICHAELMUNAVU83: I have set t in config.exs

lib/animina/narratives/resources/story.ex
MICHAELMUNAVU83: I will have to make this rule change in another PR

MICHAELMUNAVU83:
@wintermeyer  , I have added time validations for the registration_completed_at attribute well and  I have set the minimum number of stories in config.exs . 
I will work on making the `About me` story non deletable in a new PR

wintermeyer (approved):

PR #950, opened 2024-08-22
prompt update and UI fixes, Closes #940

### LINKED ISSUE #940: Fix the AI initial chat message options.
- These options have a `[]` around them. It doesn't happen every time but it doesn't make sense. Please remove the `[]` when it happens.
- The options should have a visible `list-disc` at the beginning. See https://tailwindcss.com/docs/list-style-type
- Replace `uses this` with `Copy`

Bildschirmfoto 2024-08-21 um 20 14 28

Kimutai01:
Screenshot 2024-08-22 at 14 43 59

Fixed the prompt and UI issue

### DIFF
--- a/lib/animina/chat_completion.ex
+++ b/lib/animina/chat_completion.ex
@@ -33,6 +33,8 @@ I do not need any other information.
 
 Please return the messages as follows:
 
+Do not use square brackets [] around the messages.
+
 \n\nMessage: [Message content]
 
 \n\nMessage: [Message content]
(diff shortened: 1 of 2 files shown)


lib/animina/chat_completion.ex
```diff
@@ -33,6 +33,8 @@ I do not need any other information.
 
 Please return the messages as follows:
 
+Do not use square brackets [] around the messages.
```
wintermeyer, line 36: Nice!

wintermeyer (approved):

PR #960, opened 2024-08-23
No need with your first message when answering a message , closes #952

### LINKED ISSUE #952: No need with your first message when answering a message
The help with an initial message should only be offered for the very first initial message. Not for the answer.

Bildschirmfoto 2024-08-22 um 17 43 31

MICHAELMUNAVU83:
Screenshot 2024-08-23 at 08 03 49


lib/animina_web/live/chat_live.ex
```diff
@@ -573,7 +573,7 @@ defmodule AniminaWeb.ChatLive do
           </div>
         </.form>
 
-        <%= if messages_sent_to_a_user_by_sender(@sender.id , @messages) == 0 && @current_user_credit_points > 20 && @suggested_messages == [] && @show_use_ai_button do %>
+        <%= if Enum.count(@messages) == 0 && @current_user_credit_points > 20 && @suggested_messages == [] && @show_use_ai_button do %>
```
wintermeyer: Please don't hardcode those 20 points as a fixed number! Nowhere! Please use a variable for that which gets set in config and can be overridden in dev and prod.

Please fix it for this 20 and search for other occurrences and fix them in an other PR.

BTW: It should be > 19 and not 20.

wintermeyer (changes requested):

lib/animina_web/live/chat_live.ex
MICHAELMUNAVU83: I have set this up and added an issue for it

MICHAELMUNAVU83:
@wintermeyer  hello , is there any extra change you needed on this PR?

config/config.exs
```diff
@@ -19,6 +19,9 @@ config :animina, env: Mix.env()
 # Configures the number of stories required for complete registration
 config :animina, :number_of_stories_required_for_complete_registration, 1
 
+# Configures the minimum number of points required to view the 'AI Help' Chat Message in the chat
+config :animina, :minimum_points_required_for_ai_help_in_chat, 19
```
wintermeyer: Isn't this the wrong approach? We charge 20 points for AI help for the first message, don't we? Where do we charge those 20 points? That place should use a `ai_message_help_price` which gets set in `config.exs` to 20. AND that same variable should be used in `chat_live.ex`.

- One central place to configure the value.
- Name that makes sense for somebody who reads the code for the first time.

wintermeyer (changes requested):

config/config.exs
wintermeyer: @MICHAELMUNAVU83 Just a reminder that this is not finished yet.

MICHAELMUNAVU83: @wintermeyer  , just to get it right would the only change be to change the name of the config?

wintermeyer: Not only the name. The value too. I understand why you used 19 but for any other coder who is going to read this code in a year 20 would make more sense. So please use 20 and change the other code to accommodate this.

MICHAELMUNAVU83: @wintermeyer  , I have now made those changes

wintermeyer (approved):

PR #964, opened 2024-08-24
Added Check to ensure about me is not deletable

MICHAELMUNAVU83:
- About me story can now not be deleted

### DIFF
--- a/lib/animina/validations/delete_about_story.ex
+++ b/lib/animina/validations/delete_about_story.ex
@@ -4,7 +4,8 @@ defmodule Animina.Validations.DeleteAboutStory do
   alias Animina.Narratives.Story
 
   @moduledoc """
-  This is a module for validating the first story is 'About me'
+  This is a module for ensuring the 'About me' story is not
+  deleted at all
   """
 
   @impl true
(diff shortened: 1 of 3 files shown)


lib/animina/validations/delete_about_story.ex
```diff
@@ -22,14 +23,13 @@ defmodule Animina.Validations.DeleteAboutStory do
 
     user_id = Ash.Changeset.get_attribute(changeset, opts[:user])
 
-    stories = get_stories_for_user(user_id)
-
+  
     case Headline.by_id(headline_id) do
       {:ok, headline} ->
-        if length(stories) == 1 && headline.subject == "About me" do
+        if  headline.subject == "About me" do
```
wintermeyer, line 29: This might be controversial but I think that `case` feels a bit more Elixir like than `if`.

```elixir
      case  headline.subject do
        "About me" -> 
          {:error,
           field: opts[:attribute],
           message: "You cannot delete the 'About me' story."}
        _ -> 
          :ok
        end
```

wintermeyer (approved):
I accept the PR but please change this to a `case` and use `case` more often in those scenarios.

PR #987, opened 2024-08-30
phototags resource

Kimutai01:
I decided to break the task into parts.
I created a phototag resource with a description, user and photo
You can check and advice 
Should I add a relation to a flag?


wintermeyer (changes requested):
Please replace `Tag` with `Flag`. My bad! Sorry!

It doesn't make sense to have too different words for the same thing in one application.

Kimutai01:
@wintermeyer I have updated the resource to PhotoFlagTags 
I have also added the relation to a flag

wintermeyer (changes requested):
Please rename the resource to `PhotoFlags`. Keep it simple. And please don't forget to rename the file names too. That resource e.g. was still in `photo_tags.ex`.

wintermeyer (approved):

PR #1021, opened 2024-09-05
Added country dropdown when registering , closes #963

### LINKED ISSUE #963: Country drop down box
We have to make clear that currently only German addresses are supported. Of course this is probably going to change but it is so right now. Please add a `country` attribute to `BasicUser` and `User` and resort the attributes on the index registration page like this:

- username
- password
- birthday (make that smaller)
- country (fits nicely next to birthday)
- username
- name
- the rest as it is

The `country` should be a drop down list with just one value (Germany).

NOTE: Don't forget a validation that Country can not be nil and has to be in a list ['Germany'] (so that we can enlarge that list later).

IMPORTANT: Write a migration which sets the already existing users to `Germany`.

MICHAELMUNAVU83:
- I added country dropdown that only has Germany
- I added a country attribute to `User` and `BasicUser` that has validations to ensure the value is only "Germany"
Screenshot 2024-09-05 at 14 12 47


wintermeyer (approved):
We have to change the order of the input fields. I’ll create a different issue for that.

PR #1024, opened 2024-09-07
Added Tests for non about me redirects , closes #975

### LINKED ISSUE #975: Write tests for the redirects on the router if a user has no about me story
@wintermeyer  , curious , If a user has an about me story , but has not uploaded the profile photo yet , should we allow them to use the system(although limited) , as they are not fully registered yet , or should we now redirect them to the upload profile photo page every time they log in?

MICHAELMUNAVU83:
- I added tests to ensure users cannot access routes eg , stories , dashboard , posts , chats without an about me story.
- I also changed what we had  in the router to ensure they **CAN** Change thier visibility and delete their accounts even if someone has no about story , figured this may help a user who signed up by mistake and wants to go incognito or delete their account.

### DIFF
--- a/lib/animina_web/router.ex
+++ b/lib/animina_web/router.ex
@@ -55,6 +55,8 @@ defmodule AniminaWeb.Router do
       live "/my/flags/green", FlagsLive, :green
       live "/my/flags/red", FlagsLive, :red
       live "/my/about-me", StoryLive, :about_me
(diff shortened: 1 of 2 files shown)


wintermeyer:
> I also changed what we had in the router to ensure they CAN Change thier visibility and delete their accounts even if someone has no about story

Good thinking!

PR #1035, opened 2024-09-11
Added Translations in Root Live 

MICHAELMUNAVU83:
- @wintermeyer  , we have these when I switch between german and english on my machine , be advsised that sometimes I have noticed it does not pick up the right language , Can We consider having a switch for language on the system? , maybe on the settings

Bildschirmfoto 2024-09-11 um 09 17 52
Uploading Bildschirmfoto 2024-09-11 um 09.17.57.pn
![Bildschirmfoto 2024-09-11 um 09 18 11
Screenshot 2024-09-11 at 08 56 53

Bildschirmfoto 2024-09-11 um 09 08 21

### DIFF
--- a/lib/animina_web.ex
+++ b/lib/animina_web.ex
@@ -54,6 +54,8 @@ defmodule AniminaWeb do
       use Phoenix.LiveView,
         layout: {AniminaWeb.Layouts, :app}
 
+      import Gettext, only: [with_locale: 2]
+
       unquote(html_helpers())
     end
   end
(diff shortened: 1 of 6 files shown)


MICHAELMUNAVU83:
I have sent you a mail suggesting we add a switch at the top , this can be 2 small flags in  a circle , one for US , other one German , we highlight the current language the system picks up then allow the user to switch . This will also make it easy for us to switch langauges and test out the translation as opposed to changing the computer's language preferences which will probably need a restart

wintermeyer:
> I have sent you a mail suggesting we add a switch at the top , this can be 2 small flags in a circle , one for US , other one German

@MICHAELMUNAVU83 Please do so. But don't put it super visible on every page. That would be a waste of real estate. Put it somewhere in the pull down menu. Something profile related.

PR #1039, opened 2024-09-11
test fixes, closes #939

### LINKED ISSUE #939: Add AI (LLaVA) image tagging and description
ANIMINA users add photos to their stories but we don't use those photos as input. Example: If somebody uploads a couple of photos with dogs then we can use that information to assume that this person is a dog lover. We can add the `dog` tag/flag to his/her profile. 

On production you find the following Llava models:

```text
NAME       	ID          	SIZE  	MODIFIED    
llava:13b  	0d0eb4d7f485	8.0 GB	12 days ago	
llava:7b   	8dd30f6b0cb1	4.7 GB	12 days ago
```

Please read https://ollama.com/blog/vision-models for more information about Llava. Let's use `llava:7b` for development and `llava:13b` for production.

## Tasks

- [ ] Create a new `PhotoTags` resource which links a photo to a user and tags.
- [ ] When ever a photo is created or updated run it through Llava. Give it the photo and a list of all known system tags. Plus give it a list of the white tags and tell the LLaVA that those are things the photo owner likes. Ask the LLaVA to generate a list of tags which describe the photo but only use system known tags.
- [ ] While at it generate a `description` of the photo and store it in a new attribute for that photo/image. We can use that as helpful input for #934 

Please remember that this must not at any time lead to some sort of major performance bottleneck. Think about a way to run it after a photo has been uploaded in the background. And only run a limited number of these at the same time. Please think about a solution and describe it first in this issue before implementing it.

Maybe https://github.com/thmsmlr/instructor_ex is a good approach here. But I don't know.

Kimutai01:
-Adding description and tags from a photo using llava
Screenshot 2024-09-11 at 21 05 42


Kimutai01:
@wintermeyer 
quick question,
Won't it be better to spawn the process?

wintermeyer:
> @wintermeyer quick question, Won't it be better to spawn the process?

I do not know. Believe it or not I am not that deep in Elixir to understand what is best here.

But we do have to think about overloading the whole system with this. We have to find a way to set a maximum of images which are processed in parallel. Otherwise this can become a denial of service problem.

PR #1047, opened 2024-09-15
Remove incognito option from GUI ,closes #1044

### LINKED ISSUE #1044: Remove Incognito Option
Bildschirmfoto 2024-09-13 um 08 33 56

I don't know how this made it to production. Please remove the `incognito` option here. A normal user can not make him-/herself invisible.

PS: Do we have policies to check for this? We should. Otherwise one can do it via an API call.

MICHAELMUNAVU83:
Screenshot 2024-09-15 at 14 57 19


wintermeyer (changes requested):
- This PR mixes several things. Please split it into separate PRs. One per problem/topic.
- Write a migration that takes care of accounts which have the state `incognito` and hibernate them. Please think about these things even if I don't mention them and add them by yourself.

MICHAELMUNAVU83:
@wintermeyer  , I have added the migration and trimmed the PR down to work purely on the incognito stuff

priv/repo/migrations/20240911175040_migrate_resources14.exs
```diff
@@ -0,0 +1,18 @@
...
+  use Ecto.Migration
+
+  def up do
+    # Update the users' state from 'incognito' to 'hibernate'
+    execute "UPDATE users SET state = 'hibernate' WHERE state = 'incognito';"
+  end
+
+  def down do
+    # Revert the changes, setting 'hibernate' back to 'incognito'
+    execute "UPDATE users SET state = 'incognito' WHERE state = 'hibernate';"
```
wintermeyer: This `down` migration doesn't make sense. We can not force all `hibernate` states into `incognito` back. This is a case where we have a good `up` but no `down` migration. Please remove the `down` function.

wintermeyer (changes requested):

priv/repo/migrations/20240911175040_migrate_resources14.exs
MICHAELMUNAVU83: I have removed the down function

wintermeyer (approved):

Issue #677, opened 2024-06-19
key :username not found in: nil


wintermeyer:
I did a 

```bash
$ mix ecto.reset
$ mix seed_demo_system
$ iex -S mix phx.server
```

And when I want to open a profile I get this error:

<img width="1280" alt="Bildschirmfoto 2024-06-19 um 09 43 20" src="https://github.com/animina-dating/animina/assets/69770/8fe7bd8d-74b7-45e1-9a00-199376bf0bf2">

PS: Please write a test for this too.

MICHAELMUNAVU83:
@wintermeyer  , did you access this as an anonymous user?

wintermeyer:
> @wintermeyer , did you access this as an anonymous user?

Yes, I did.

MICHAELMUNAVU83:
Okay , the test is actually there , these problems occurs if the session is cached , which happens when you run the reset commands , and we cannot have tests for cached sessions. We will have to sort them out as they come.

PR #1050, opened 2024-09-16
remove redundant mount/3 in profile_live, closes #1049

### LINKED ISSUE #1049: Refactor the mount/3 functions in profile_live

briankariuki:
I've removed some redundant mount/3 code as its covered by the profile_socket functions. I've tested it both logged in and logged out and it works.

### DIFF
--- a/lib/animina_web/live/profile_live.ex
+++ b/lib/animina_web/live/profile_live.ex
@@ -312,11 +312,6 @@ defmodule AniminaWeb.ProfileLive do
     end
   end
 
-  defp redirect_to_username(socket, username) do
-    socket
-    |> push_navigate(to: ~p"/#{username}")
-  end
-
   defp subscribe(socket, current_user, user) do
     if connected?(socket) do
       PubSub.subscribe(Animina.PubSub, "credits")
--- a/lib/animina_web/live/profile_live.ex
+++ b/lib/animina_web/live/profile_live.ex
@@ -237,6 +237,19 @@ defmodule AniminaWeb.ProfileLive do
     end
   end
 
+  def mount(_params, %{"language" => _language}, socket) do
+    current_user =
+      socket.assigns.current_user
+
+    if current_user do
+      {:ok,
+       socket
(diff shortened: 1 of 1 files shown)


MICHAELMUNAVU83:
@briankariuki  , This was there for if someone logs out but there is a cache , so the system thinks there is a `current_user` but there is none , it broke the system , that is why it was there .

briankariuki:
> @briankariuki , This was there for if someone logs out but there is a cache , so the system thinks there is a `current_user` but there is none , it broke the system , that is why it was there .

How can I recreate this scenario?

MICHAELMUNAVU83:
That was the problem actually , it is a sort of scenario we cannot replicate or test for , it happens sometimes when a cache happens , might you know how to simulate a cache

MICHAELMUNAVU83:
@briankariuki  , here is the issue https://github.com/animina-dating/animina/issues/677

wintermeyer:
> @briankariuki , here is the issue #677

With that information @briankariuki can replicate the problem. Build one system. Log in. Reset the database. Et voila!

wintermeyer:
But: One can argue that this problem never occurs in the real world since we don't reset production every day. So I would be open to accept this problem since it is only relevant on development systems.

briankariuki:
> But: One can argue that this problem never occurs in the real world.

True. I think we can close this

wintermeyer (approved):
@briankariuki Please do test this in production when we deploy it and have an eye on it 3-4 days after deployment. I don't want to be surprised by a bug we didn't see coming.

wintermeyer:
@briankariuki I just triggered a deployment to production. You can see the time of the last commit of a production version in the footer (it is encoded in the version number).

briankariuki:
> @briankariuki I just triggered a deployment to production. You can see the time of the last commit of a production version in the footer (it is encoded in the version number).

Will check in a few days

wintermeyer:
> Will check in a few days

Please do check today too.

Issue #1002, opened 2024-09-02
Remove CategoryTranslation and FlagTranslation


wintermeyer:
It seemed to be a good idea at the time but `CategoryTranslation` and `FlagTranslation` are not needed. We just display the values with `gettext/1` and use the I18n mechanisms to translate them. Please remove those resources from the code and write migrations which delete them from the database.

Bonus: It would be create to save the German translations in a `.po` file so that we don't have to translate them again.

## Tasks
- [x] #1071
- [x] #1072
- [x] #1073
- [x] #1076

briankariuki:
@wintermeyer Did you manage to create the po file for german translations?

wintermeyer:
> @wintermeyer Did you manage to create the po file for german translations?

I mispelled that part of the issue. What I meant: It would be great if you took the content of the SQL database and create a .po file with it for the German translation.

wintermeyer:
@briankariuki But don't do it in case it is a lot of work. We can use ChatGPT to translate the English .po file to a German one.

PR #1057, opened 2024-09-18
add action to get intersecting flags, closes #1056

### LINKED ISSUE #1056: Add action to user_flags that loads the intersecting flags between two users based on color

briankariuki:
I have added an action which uses only 1 sql query to get intersecting flags for two users based on color.
The action uses Ecto.Query instead of Ash.Query, enabling subqueries (to load flag and flag_translations), joins and grouping at the data_layer.
Suggestions to improve this action are welcome.

### DIFF
--- a/lib/animina/traits/resources/user_flags.ex
+++ b/lib/animina/traits/resources/user_flags.ex
@@ -1,6 +1,6 @@
 defmodule Animina.Traits.UserFlags do
-  alias Animina.Validations
   alias Animina.Repo
+  alias Animina.Validations
 
   import Ecto.Query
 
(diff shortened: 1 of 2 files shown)


wintermeyer:
> which uses only 1 sql query to get intersecting flags for two users

Just out of curiosity: How many SQL queries did we need before?

wintermeyer (approved):
Great work!

Please just fix the credo stuff.

briankariuki:
> > which uses only 1 sql query to get intersecting flags for two users
> 
> Just out of curiosity: How many SQL queries did we need before?

number of user_flags x 2(1 for flag, 1 for flag_translation) per user. 

And these would be called a few times in the profile_live mount

wintermeyer:
> And these would be called a few times in the profile_live mount

👍

Next step: We have to get rid of those translations. See https://github.com/animina-dating/animina/issues/1002

briankariuki:
> Great work!
> 
> Please just fix the credo stuff.

For some reason running `mix credo --strict` results in an error

image

wintermeyer:
> For some reason running `mix credo --strict` results in an error

Strange. It runs through in GitHub.

PR #1091, opened 2024-10-10
Add fast user resource, closes #1092

### LINKED ISSUE #1092: implement Fast User resource

briankariuki:
This resource is similar to the user resource, but it loads all calculations (city, roles, profile_photo) and aggregates (credit_points) in one sql query.

### DIFF
--- a/lib/animina/accounts/resources/fast_user.ex
+++ b/lib/animina/accounts/resources/fast_user.ex
@@ -199,88 +199,31 @@ defmodule Animina.Accounts.FastUser do
 
   attributes do
     uuid_primary_key :id
-    attribute :email, :ci_string, allow_nil?: false, public?: true
-    attribute :hashed_password, :string, allow_nil?: false, sensitive?: true
+    attribute :email, :ci_string, public?: true
     attribute :is_in_waitlist, :boolean, default: false, public?: true
-
-    attribute :username, :ci_string do
-      allow_nil? false
-      public? true
-
-      constraints max_length: 15,
-                  min_length: 2,
-                  match: ~r/^[A-Za-z0-9._-]*$/,
-                  trim?: true,
-                  allow_empty?: false
-    end
-
-    attribute :name, :string do
-      allow_nil? false
-      public? true
-
-      constraints max_length: 50,
-                  min_length: 1,
-                  trim?: true,
-                  allow_empty?: false
-    end
-
-    attribute :birthday, :date, allow_nil?: false, public?: true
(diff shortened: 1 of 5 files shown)


lib/animina/accounts/resources/fast_user.ex
```diff
@@ -0,0 +1,116 @@
...
+  This is the Fast User module.
+  """
+
+  use Ash.Resource,
+    data_layer: AshPostgres.DataLayer,
+    authorizers: Ash.Policy.Authorizer,
+    domain: Animina.Accounts,
+    extensions: []
+
+  postgres do
```
wintermeyer, line 23: Is it possible to set this read only? If yes we should.

lib/animina/accounts/resources/fast_user.ex
```diff
@@ -0,0 +1,116 @@
...
+      run fn input, _ ->
+        {:ok, []}
+      end
+    end
+  end
+
+  attributes do
+    uuid_primary_key :id
+    attribute :email, :ci_string, allow_nil?: false, public?: true
+    attribute :hashed_password, :string, allow_nil?: false, sensitive?: true
```
wintermeyer: Let's keep this to the minimum which we need for the profile view. Which means e.g. no need for the `:hashed_password` and other attributes. Strip it down as fare as possible.

lib/animina/accounts/resources/fast_user.ex
```diff
@@ -0,0 +1,116 @@
...
+    attribute :is_in_waitlist, :boolean, default: false, public?: true
+
+    attribute :username, :ci_string do
+      allow_nil? false
```
wintermeyer: If we make the resource read-only we don't need the constraints or validations.

wintermeyer (changes requested):

lib/animina/accounts/resources/fast_user.ex
briankariuki: I couldn't find any documentation on how to make a resource read only. FastUser only supports the read action

wintermeyer (approved):
Nice work @briankariuki !

Maybe we need a bit more documentation for other developers to understand the code but this is a major speed improvement. I like it.

briankariuki:
> Nice work @briankariuki !
> 
> Maybe we need a bit more documentation for other developers to understand the code but this is a major speed improvement. I like it.

I'll add explanations in the discussion tab later in the week

PR #1096, opened 2024-10-14
fix delete photo bug, closes #1093

### LINKED ISSUE #1093: Old pictures need to be deleted from the database
Pictures which aren't in use anymore are still accessible via deep link.
They should be deleted from the system instead.
AFAIK this might be necessary according to EU law.

Kimutai01:
-Fixed the delete photo bug for all photos including about me. The photos are deleted from both the database and the file system.

https://github.com/user-attachments/assets/705ec7ad-205c-44b4-b3d4-de609018a97a

### DIFF
--- a/lib/animina/accounts/resources/photo.ex
+++ b/lib/animina/accounts/resources/photo.ex
@@ -7,11 +7,12 @@ defmodule Animina.Accounts.Photo do
   alias Animina.Accounts.PhotoFlags
   alias Animina.ImageTagging
   alias Animina.Narratives
-  alias Animina.Validations
 
   Animina.Validations.AboutPhoto
 
   alias Animina.Traits.Flag
+
+  require Ash.Query
 
   require Logger
 
@@ -197,9 +198,9 @@ defmodule Animina.Accounts.Photo do
            on: [:create, :destroy]
   end
 
-  validations do
-    validate {Validations.AboutPhoto, story: :story_id, user: :user_id}, on: :destroy
-  end
+  # validations do
+  #   validate {Validations.AboutPhoto, story: :story_id, user: :user_id}, on: :destroy
+  # end
 
   attributes do
     uuid_primary_key :id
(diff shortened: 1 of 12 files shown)


lib/animina_web/live/profile_photo_live.ex
```diff
@@ -115,7 +115,7 @@ defmodule AniminaWeb.ProfilePhotoLive do
         params = Map.merge(params, file)
         form = Form.validate(socket.assigns.form, params, errors: true)
 
-        Photo.destroy(socket.assigns.current_user.profile_photo)
+        Photo.delete_photo_and_optimized_photos(socket.assigns.current_user.profile_photo)
```
wintermeyer: This solves the problem on the wrong layer of the application. What if I open the `iex` and delete a user manually? Or a photo resource? We need to solve this in the resource and not in the LiveView layer.

wintermeyer (changes requested):
Always solve problems in a way that it works for an API call too or in the `iex`. Not just the one website you are working on. There is always the possibility that an other developer creates a similar page but forgets or doesn't know about this and there for doesn't think about deleting the files. Please move the functionality to resource level.

Kimutai01:
Uploading Untitled design (81).mp4…

lib/animina_web/live/profile_photo_live.ex
wintermeyer: If you were new to this project and wanted to delete a `Photo`, would you use `destroy` or `delete_photo_and_optimized_photos`? Rewrite the `destroy` action so that we can use it here as it feels natural. We `destroy` a `Photo` and the resource takes care of all the details.

wintermeyer (changes requested):

Kimutai01:
@wintermeyer 
My bad..It was a mistake from my side.
Right now the files are deleted after destroy

https://github.com/user-attachments/assets/784741a4-bc93-4439-b936-b48f25975f9e

lib/animina/accounts/resources/photo.ex
```diff
@@ -195,12 +196,19 @@ defmodule Animina.Accounts.Photo do
...
-    validate {Validations.AboutPhoto, story: :story_id, user: :user_id}, on: :destroy
+    change after_action(fn changeset, record, _ ->
+             delete_photo_and_optimized_photos(record)
+
+             {:ok, record}
+           end),
+           on: [:destroy]
   end
 
+  # validations do
```
wintermeyer: Why don't we have validations? How can we ensure that a `story` exists and that the `story_id` is set?

wintermeyer (changes requested):

Kimutai01:
@wintermeyer 
I have updated the validations.
I had forgotten this. I got into a rabbit hole of the image AI implementation in the 2 issues😅.

Here is the new video after the updates.

https://github.com/user-attachments/assets/87ab24c3-83f5-4d2e-a76c-89efe1e2ddf4

wintermeyer (approved):

Issue #1048, opened 2024-09-16
Creating/Updating a story takes for ever


wintermeyer:
Please have a look into this problem. It takes forever to create or update a story. At one time at the end we even get an error message. This is all on production.

![ScreenFlow](https://github.com/user-attachments/assets/0460c290-55d0-49c4-aa0b-b96a32710f35)

Please find out the problem and fix it.

MICHAELMUNAVU83:
This is most probably caused by the photo tagging bit by AI , is it possible to spawn that function @Kimutai01

Kimutai01:
@MICHAELMUNAVU83 
We can spawn the function. I also realized it was timing out at times on localhost but I thought it was because I was running the model locally,
but I'm looking into it

wintermeyer:
We do not need the tags right away but we need the story to be saved as soon as possible. So any kind of queuing mechanism would work fine too. Something that is robust enough to survive a) a new deployment and b) a reboot of the server.

wintermeyer:
>  I also realized it was timing out at times on localhost but I thought it was because I was running the model locally

ALWAYS test this on production after deployment too. It must not be the case that I am the only one who catches these kind of problems/bugs.

Kimutai01:
I

> > I also realized it was timing out at times on localhost but I thought it was because I was running the model locally
> 
> ALWAYS test this on production after deployment too. It must not be the case that I am the only one who catches these kind of problems/bugs.

It's okay I'll always test it both on production and locally.
I'm working on it ASAP

briankariuki:
@Kimutai01 @MICHAELMUNAVU83 I also think there's some refactoring opportunity here. after_action hooks in my opinion should not have any expensive functions attached. The `create_optimized_photos/1` and `create_photo_flags/1` may take time to execute

```elixir

change after_action(fn changeset, record, _ ->
   create_optimized_photos(record)
   create_photo_flags(record)
   #  spawn(fn -> create_photo_flags(record) end)

   {:ok, record}
 end),
 on: [:create]

```

MICHAELMUNAVU83:
Which is the best way to handle such an operation?

briankariuki:
> Which is the best way to handle such an operation?

Use some task runner like oban, or a genserver/genstage setup.

wintermeyer:
> Use some task runner like oban, or a genserver/genstage setup.

[...]

PR #1105, opened 2024-10-20
photo tag genstage, closes, #1048

Kimutai01:
-I have created a  genserver, genstage setup to queue the images for tagging. I also added an attribute in the resource called tagged which is updated when the image is tagged.


lib/animina/accounts/resources/photo.ex
```diff
@@ -311,9 +315,10 @@ defmodule Animina.Accounts.Photo do
 
   def create_photo_flags(record) do
     if File.exists?("priv/static/uploads/" <> record.filename) do
-      {flags, description} = ImageTagging.tag_image_using_llava("#{record.filename}")
+      {flags, description} =
+        ImageTagging.tag_image_using_llava("#{record.filename}")
```
wintermeyer: I don't like the function name "tag_image_useing_llava". Mainly because it hs `llava` hardcoded in it. We can change the LLM any day and we will (probably within 12 months). Keep it generic. Something like `auto_tag_image`.

lib/animina/accounts/resources/photo.ex
```diff
@@ -311,9 +315,10 @@ defmodule Animina.Accounts.Photo do
...
 
-      case update(record, %{description: description}) do
+      case update(record, %{description: description, tagged: true}) do
```
wintermeyer: Maybe I don't understand this right but why do we update to a `tagged: true` here when the image is already tagged by `tag_image_using_llava`? Shouldn't that function auto update that attribute right after tagging the resource?

wintermeyer (changes requested):
Before requesting a review again please send an email to @wintermeyer describing how the system handles a power outage. Assuming we have 100 images in the pipeline.

lib/animina/accounts/resources/photo.ex
Kimutai01: @wintermeyer 
What I'm doing here is once we get a response from the llm, I update the description and set tagged. I can also do it using in the after action

wintermeyer (changes requested):
Please rename tagged to tagged_at and make it a timestamp. 

Why? We will change LLMs in the future. Than we will rerun the tagging. For that the timestamp is useful.

wintermeyer (approved):

PR #1110, opened 2024-11-02
Replaced Confirmation Link with Confirmation PIN that has 3 max attempts

MICHAELMUNAVU83:
https://github.com/user-attachments/assets/696122db-dca1-49e0-869f-b83a435309a3

Screenshot 2024-11-02 at 08 27 42
Screenshot 2024-11-02 at 08 37 35

@wintermeyer  , I have added the Confirmation PIN to replace the link , it has a maximum of 3 attempts after which we delete accounts , as you are the master of content , kindly advise on how the email validation and pin page is and how that email should look like

**Features**

- Users can resend the email with the pin
- Users can have 3 retries
- There is a counter on how many retries are left
- On the last retry , there is a red box telling the user they have one attempt


lib/animina/accounts/resources/user.ex
```diff
@@ -201,7 +191,9 @@ defmodule Animina.Accounts.User do
         :confirmed_at,
         :is_in_waitlist,
         :registration_completed_at,
-        :country
+        :country,
+        :confirmation_pin,
```
wintermeyer, line 195: I don't like the idea of having the `confirmation_pin` stored in clear in the database. Please store a secure hash sum and compare that with the entered PIN.

lib/animina/accounts/resources/user.ex
```diff
@@ -503,6 +496,12 @@ defmodule Animina.Accounts.User do
     attribute :minimum_partner_height, :integer, allow_nil?: true
     attribute :maximum_partner_height, :integer, allow_nil?: true
 
+    attribute :confirmation_pin, :integer do
```
wintermeyer: This shouldn't be able to be nil. The PIN or better the hash should be created during the create action. If I create a new user in the iex I want it to have a PIN/hash automatically.

Actually the `confirmation_pin` should be a virtual field which can be written to when I create a user in the iex. That input should be automatically converted to the hash and stored in the hash.

lib/animina/accounts/resources/user.ex
```diff
@@ -621,6 +620,14 @@ defmodule Animina.Accounts.User do
     end
   end
 
+  defp add_confirmation_pin_to_user_and_send_email(record) do
+    random_pin = :rand.uniform(900_000) + 100_000
```
wintermeyer: Nice touch that the PIN can not start with a 0. Good thinking!

lib/animina/user_email.ex
```diff
@@ -29,20 +27,19 @@ defmodule Animina.UserEmail do
...
-
-    Your new ANIMINA https://animina.de account has been created with this
+  defp construct_email_body(user) do
+    "Your new ANIMINA https://animina.de account has been created with this
     email address.
 
-    Please confirm the email address by clicking the monstrosity of a link
-    below or do nothing in case you didn't create the account.
+    Below is your super secret 6 digit pin that you can use to verify your account once you
+    visit  https://animina.de/my/email-validation , you have only 3 attemps so make sure
```
wintermeyer: A user might create a new account on his desktop but reads this email on his mobile phone. Clicking on the URL in the mobile phone will get him nowhere.

Better text:
Please use this 6 digit PIN to verify your new ANIMINA account:

wintermeyer (changes requested):

wintermeyer:
`Hi test@gmail.com!`

Can't we do better? We already now the name of this human, don't we?

lib/animina/accounts/resources/user.ex
MICHAELMUNAVU83: @wintermeyer  hello , if we hash with bcrypt and store it a pin that is "123456" as "$2b$12$uf/Ow4LXf54uGjo5A3WrP.aWOKDEHQu3zCVbFBgwKmQ8fbg837xLm" , knowing there is no way to get the value from a hash , we can only compare 2 values , how would we send the email to the user with this pin , keep in mind too , on that page they can request for another email to be sent , so we have to store that value explicitly some way , thoughts?

lib/animina/user_email.ex
MICHAELMUNAVU83: This has been done , thanks

lib/animina/accounts/resources/user.ex
MICHAELMUNAVU83: Okay ,  kindly check [ Bcypt Documentation]( https://hexdocs.pm/bcrypt_elixir/Bcrypt.html) and  the comment below and advise , We need a way to store it to be able to send it to the user if they request for additional emails

lib/animina/accounts/resources/user.ex
wintermeyer: We write a `set_new_pin_and_send_by_email` function which does exactly that. It generates a new PIN, stores the hash, resets the counter and sends the email to the user. The PIN is only known within that function. When a user requests a second email he gets a new PIN.

BTW: We want to include the UserID in the hash too. Otherwise it would be easy to compare to users who have the same PIN.

wintermeyer: The `set_new_pin_and_send_by_email` function is called in the `create` action.

MICHAELMUNAVU83: Okay , that does make sense , I have set it as such  we have this in the before create action , so we cannot access ther user_id in this case

MICHAELMUNAVU83:
@wintermeyer  , I worked on the  changes you requested

lib/animina/accounts/resources/user.ex
```diff
@@ -417,6 +408,29 @@ defmodule Animina.Accounts.User do
            end),
            on: [:create]
 
+    change before_action(fn changeset, record ->
+             random_pin =
+               (:rand.uniform(900_000) + 100_000)
+               |> Integer.to_string()
+
+             hashed_pin =
+               random_pin
```
wintermeyer: You need to "salt" the PIN (see https://crackstation.net/hashing-security.htm). hmmm... normally I would have used the `user.id` but here we don't have that yet or do we? Let's use the email address as a salt.

lib/animina/accounts/resources/user.ex
```diff
@@ -417,6 +408,29 @@ defmodule Animina.Accounts.User do
            end),
            on: [:create]
 
+    change before_action(fn changeset, record ->
+             random_pin =
```
wintermeyer: I understand why you used `random_pin` as a variable name and it is not a bad choice. But during the rest of the code I stumble upon it. `new_pin` is better. It is a small thing but it helps me to better understand the code.

lib/animina/user_email.ex
```diff
@@ -6,43 +6,38 @@ defmodule Animina.UserEmail do
   import Swoosh.Email
   import AniminaWeb.Gettext
   alias Animina.Mailer
-  use AshAuthentication.Sender
 
-  def send(user, confirm, _opts) do
+  def send_pin(name, email, pin) do
     subject = gettext("👫❤️ Confirm the email address for your new ANIMINA account")
```
wintermeyer: Not bad but I think a short subject line is better:
"👫❤️ Confirm your ANIMINA account"

lib/animina/user_email.ex
```diff
@@ -6,43 +6,38 @@ defmodule Animina.UserEmail do
...
-    ~S"""
-
-    Your new ANIMINA https://animina.de account has been created with this
+  defp construct_email_body(pin) do
+    "Your new ANIMINA https://animina.de account has been created with this
     email address.
 
-    Please confirm the email address by clicking the monstrosity of a link
-    below or do nothing in case you didn't create the account.
+    Please use this 6 digit PIN to verify your new ANIMINA account.
```
wintermeyer: Please add:
\n
If you didn't create an account just do nothing. We will auto delete the entry within 24 hours.

lib/animina/user_email.ex
```diff
@@ -6,43 +6,38 @@ defmodule Animina.UserEmail do
...
 
-    """
-  end
 
-  defp construct_link(confirm) do
-    "\n #{"#{get_link(Application.get_env(:animina, :env))}/auth/user/confirm_new_user?#{URI.encode_query(confirm: confirm)}"}"
+    #{pin}
```
wintermeyer, line 37: Please do me a favour and look at the email and sprinkle a little bit of design/layout love over it if needed. I know that we don't use HTML here but even an ASCII only email can look nice. Do we have multiple new lines without sense? Do we indent the PIN to make it easier to spot?

lib/animina_web/live/email_validation_live.ex
```diff
@@ -11,22 +12,154 @@ defmodule AniminaWeb.EmailValidationLive do
     {:ok, socket}
   end
 
+  @impl true
+
+  def handle_event("verify_pin", %{"user" => %{"pin" => ""}}, socket) do
+    {:noreply,
+     socket
+     |> put_flash(:error, "Please enter the PIN")}
```
wintermeyer: I18n???

lib/animina_web/live/email_validation_live.ex
```diff
@@ -11,22 +12,154 @@ defmodule AniminaWeb.EmailValidationLive do
...
+
+  def handle_event("verify_pin", %{"user" => %{"pin" => pin}}, socket) do
+    if Bcrypt.verify_pass(pin, socket.assigns.current_user.confirmation_pin) do
+      correct_pin_socket(socket)
+    else
+      incorrect_pin_socket(socket)
+    end
+  end
+
+  def handle_event("resend_pin", _params, socket) do
```
wintermeyer: `resend_pin` is a misleading wording. It should be `send_a_new_pin` since we can not resend a PIN. We can just create a new one and send that.

lib/animina_web/live/email_validation_live.ex
```diff
@@ -11,22 +12,154 @@ defmodule AniminaWeb.EmailValidationLive do
...
+    if socket.assigns.current_user.confirmation_pin_attempts == 2 do
+      User.destroy(socket.assigns.current_user)
+
+      {:noreply,
+       socket
+       |> push_navigate(to: "/")
+       |> put_flash(
+         :error,
+         with_locale(socket.assigns.language, fn ->
+           gettext("Account Deleted , Kindly Sign Up Again")
```
wintermeyer: Better text:
You entered a wrong PIN three times. Therefor we deleted the account. But you can create a new one right away.

wintermeyer: Please do test if the user actually can create a new one right away. Make sure that we delete all cookies and stuff in this step.

lib/animina_web/live/email_validation_live.ex
```diff
@@ -11,22 +12,154 @@ defmodule AniminaWeb.EmailValidationLive do
...
+      User.update(socket.assigns.current_user, %{
+        confirmation_pin_attempts: socket.assigns.current_user.confirmation_pin_attempts + 1
+      })
+
+      {:noreply,
+       socket
+       |> push_navigate(to: "/my/email-validation")
+       |> put_flash(
+         :error,
+         with_locale(socket.assigns.language, fn -> gettext("Wrong PIN Entered !") end)
```
wintermeyer: "Wrong PIN Entered!"

lib/animina_web/live/email_validation_live.ex
```diff
@@ -11,22 +12,154 @@ defmodule AniminaWeb.EmailValidationLive do
...
+     socket
+     |> push_navigate(to: "/my/potential-partner")
+     |> put_flash(
+       :info,
+       with_locale(socket.assigns.language, fn -> gettext("Account Confirmed Successfully") end)
+     )}
+  end
+
+  defp set_new_pin_and_send_by_email(user) do
+    random_pin =
```
wintermeyer: I am not a big fan of the duplication of this code since we have something similar at a different place. Please DRY it so that we have ONE function in the project which generates a new PIN and a hash for that. 

Actually please create ONE function `generate_pin_and_email_it(user)` which does it and which can be called here and at the other place.

wintermeyer (changes requested):
I am ok with it now but for the future and other features: Please don't hard code something like the length of a PIN in the code multiple times. That should be a variable which can be set in the config. Maybe we want to change it to a 4 digit or even a 8 digit PIN in the future.

lib/animina/accounts/resources/user.ex
MICHAELMUNAVU83: Yeah we can combine the 2 then hash them , that would then be "michaelmunavu83@gmail.com123456" , here "123456" is the actual pin

lib/animina/user_email.ex
MICHAELMUNAVU83: Yes , this is how it looks

Screenshot 2024-11-04 at 07 51 18

It is easy to see the pin

lib/animina_web/live/email_validation_live.ex
MICHAELMUNAVU83: I have set up gettext for this

lib/animina_web/live/email_validation_live.ex
MICHAELMUNAVU83: We can do this for when a user is generating a new pin , in that we need to generate a pin and update user resource with 'confirmation_pin' and 'confirmation_pin_attempts' , but when creating the user , we generate a pin and update the changeset with confirmation_pin only . What we can have is a common generate_pin function

MICHAELMUNAVU83: I have set it up as such for create and update actions

lib/animina_web/live/email_validation_live.ex
MICHAELMUNAVU83: I have set this up

lib/animina/user_email.ex
wintermeyer: I'd prefer something like:

```text
Hi kamala!

Please use this PIN to verify your new ANIMINA account:

    315223

Best regards,
ANIMINA Team

-- 
https://animina.de
This email was sent by the ANIMINA system. Please do not reply to it.
I am just dumb software and can't read your emails. Do contact my human boss Stefan Wintermeyer in case you have any questions or
```

wintermeyer (approved):

PR #1126, opened 2024-11-06
Refactor/fast profile closes #1109

### LINKED ISSUE #1109: Refactor top navigation bar

briankariuki:
This PR improves the FastUser query. This achieved by doing 
1. Adding indexes for table fields used in left_join query for fast_user query builder
2. Replace the six_random_public_users/1 function in top_navigation_component with an action that utilizes the FastUser query

I've tested the result on an M2 Mac, your testing and cpu may vary. Here are some before and after query times

Before adding indexes, the query times would take about 50ms
Screenshot 2024-11-06 at 09 45 19

After adding indexes, the slowest query takes about 25ms
Screenshot 2024-11-06 at 09 45 51

On further improvement to the filters used for the random interests, we now have about 10 - 15ms for the slowest queries
Screenshot 2024-11-06 at 10 02 33

Note that due to liveview's double remounting the queries fetching for random interests are executed twice. We can improve this by making the top_navigation_bar its own nested liveview, which would save the number of sql queries executed even on navigation to other routes.


wintermeyer (approved):
A lot of new indexes! But we have bigger fish to fry right now than to save space on hard drive and slower writes.

PR #1127, opened 2024-11-06
Spawned Email sending to ensure faster transition to pin  page , closes #1122

### LINKED ISSUE #1122: PIN page too slow! 😱
It takes "forever" when I click on the registration page (the first) one till the PIN page gets displayed. Why? What do we do there that takes so long? Is it possible that we wait for the email to be sent (we shouldn't - that can be done in the background).

ScreenFlow

THIS NEEDS TO BE SNAPPY! Please have to think "WOW, this page is fast!"

MICHAELMUNAVU83:
- This PR ensures we do not wait for the pin email to go out before redirecting user to the pin page , we spawn this email sending function

### DIFF
--- a/lib/animina/accounts/resources/user.ex
+++ b/lib/animina/accounts/resources/user.ex
@@ -666,7 +666,7 @@ defmodule Animina.Accounts.User do
     hashed_pin =
       hash_pin(email, new_pin)
 
-    UserEmail.send_pin(name, email, new_pin)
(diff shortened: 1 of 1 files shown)


wintermeyer (approved):

wintermeyer:
@MICHAELMUNAVU83 I just triggered a deployment run. That takes up to 5 minutes. Please test the feature later and create a new issue in case there is still a problem. It should be snappy!

PR #1141, opened 2024-11-12
Fix photo atom state bug

MICHAELMUNAVU83:
I noticed a bug where sometimes we have the state as "approved" , String , instead of :approved , atom , this PR ensures we only use the atom version of it.
Screenshot 2024-11-12 at 09 20 49

### DIFF
--- a/lib/animina_web/components/profile/stories_components.ex
+++ b/lib/animina_web/components/profile/stories_components.ex
@@ -92,9 +92,6 @@ defmodule AniminaWeb.StoriesComponents do
         </h3>
       </div>
       <div class="pb-2">
-        <p>
-          <%= @story.photo.state %>
-        </p>
         <.story_body
           story={@story}
           user={@user}
--- a/lib/animina_web/components/profile/stories_components.ex
+++ b/lib/animina_web/components/profile/stories_components.ex
@@ -17,6 +17,9 @@ defmodule AniminaWeb.StoriesComponents do
   attr :user, :any, required: false
   attr :language, :any, required: true
 
+  # we display the state as an atom in the photo struct, but we need to make sure it is an atom
+  # as we are using it for pattern matching
+
   def stories_display(assigns) do
     ~H"""
     <div class="flex flex-col gap-4">
@@ -71,6 +74,7 @@ defmodule AniminaWeb.StoriesComponents do
           }
           src={Photo.get_optimized_photo_to_use(@story.photo, :normal)}
           alt={@story.headline.subject}
+          id={"photo-for-story-#{@story.id}"}
           class="absolute inset-0 object-cover w-full h-full -z-10"
         />
 
@@ -88,6 +92,9 @@ defmodule AniminaWeb.StoriesComponents do
         </h3>
       </div>
       <div class="pb-2">
+        <p>
+          <%= @story.photo.state %>
+        </p>
         <.story_body
           story={@story}
           user={@user}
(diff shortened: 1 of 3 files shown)


lib/animina_web/components/profile/stories_components.ex
```diff
@@ -29,6 +29,13 @@ defmodule AniminaWeb.StoriesComponents do
             language={@language}
             delete_story_modal_text={@delete_story_modal_text}
             user={@user}
+            state={
```
wintermeyer, line 35: `photo.state` is not set by some magic in the background which we can not control. If there is a bug somewhere in our code which sets this `state` in a wrong way we have to fix THAT bug and not add more code afterwards to "fix" it. This shouldn't happen in the first place. Please find the original bug!

wintermeyer (changes requested):
Good: Good catch!
Bad: Wrong way of fixing it.

lib/animina_web/components/profile/stories_components.ex
wintermeyer: And (of course!) after fixing the original bug we have to make sure that already existing database entries are fixed too (e.g. with a migration).

MICHAELMUNAVU83: I thought about this , in the database with IEX , all come as an atom , also we have it in the validation

```
  attribute :state, :atom do
      constraints one_of: [:pending_review, :in_review, :approved, :rejected, :error, :nsfw]

```

So this must be an ash bug or something similar where it displays atoms in the database as strings.
Although when you have an atom for example :stefan in the HTML , it is displayed as "stefan"

wintermeyer: - Is there code somewhere which writes into the database while bypassing the validation?
- If this is an Ash bug: Did you ask in the elixirforum.com about it and/or created an issue in the Ash GitHub repo?

MICHAELMUNAVU83: 1. We use the state transition which does so implicitly , @briankariuki  , I used some of your code  
```
 <%= state =
            if(
              is_atom(@story.photo.state),
              do: @story.photo.state,
              else: String.to_atom(@story.photo.state)
            ) %>
``` , do you mind chipping in here?

wintermeyer: > Is there code somewhere which writes into the database while bypassing the validation?

My bad. I just re-read your comment. So in the DB everything is fine. hmmm... does seem like an Ash bug. Please ask in the elixirforum.com if there is an other explanation for this and add a link to that question here.

In addition: 
- Add a comment in the code which describes why you are doing this. Otherwise somebody will remove that code in the future since it shouldn't be a problem in the first place. 
- If possible write a test to make sure that we don't run into a problem if somebody deletes this code anyway.
- Resubmit the PR to be reviewed.

MICHAELMUNAVU83: I have  added tests and comments for clarity.
I have also asked a question on elixir forum for the same https://elixirforum.com/t/using-database-values-stored-as-atoms-in-html-in-ash-framework/67437

wintermeyer (approved):

MICHAELMUNAVU83:
@wintermeyer  , kindly let me know once you deploy this

wintermeyer:
> @wintermeyer , kindly let me know once you deploy this

I just triggered the deployment. ETA: +5 minutes

MICHAELMUNAVU83:
All good , I can now see your face
Screenshot 2024-11-12 at 13 59 28
Let me do a quick check for other places we use the state conditionally

MICHAELMUNAVU83:
Seems like we have a couple of similar issues related to this bug causing photos not be seen , There will be a PR for this

wintermeyer:
> Seems like we have a couple of similar issues related to this bug causing photos not be seen , There will be a PR for this

Can we add a test for this?

MICHAELMUNAVU83:
Yes

PR #1144, opened 2024-11-13
Fix state checker for images to ensure they can be seen , closes #1143

### LINKED ISSUE #1143: Images are not Visible If they Are approved
Images are not visible that are approved , this is because we have a checker that is looking for the state :appoved but ash returns it as "approved" , thus , we do not see the image

Screenshot 2024-11-13 at 06 02 33

MICHAELMUNAVU83:
Good news , we can now see images  , I also added test cases for this
Screenshot 2024-11-13 at 05 52 47
Screenshot 2024-11-13 at 05 51 32
Screenshot 2024-11-13 at 05 51 50
Screenshot 2024-11-13 at 05 52 18

### DIFF
--- a/lib/animina_web/components/bookmark/bookmark_component.ex
+++ b/lib/animina_web/components/bookmark/bookmark_component.ex
@@ -55,6 +55,17 @@ defmodule AniminaWeb.BookmarkComponent do
     end
   end
 
+  # we display the state as an atom in the photo struct, but we need to make sure it is an atom
+  # as we are using it for pattern matching
+
+  defp make_sure_photo_state_is_atom(nil) do
+    ""
+  end
+
+  defp make_sure_photo_state_is_atom(photo) do
+    String.to_atom(photo.state)
+  end
+
   @impl true
   def render(assigns) do
     ~H"""
@@ -65,6 +76,13 @@ defmodule AniminaWeb.BookmarkComponent do
         language={@language}
         reason={@reason}
         current_user={@current_user}
+        state={
+          if @current_user.profile_photo && is_atom(@current_user.profile_photo.state) do
+            @current_user.profile_photo.state
+          else
+            make_sure_photo_state_is_atom(@current_user.profile_photo)
+          end
+        }
         delete_bookmark_modal_text={
           with_locale(@language, fn -> gettext("Do you really want to delete this bookmark?") end)
         }
--- a/lib/animina_web/components/chat/chat_components.ex
+++ b/lib/animina_web/components/chat/chat_components.ex
@@ -115,7 +115,16 @@ defmodule AniminaWeb.ChatComponents do
             </div>
           </div>
         </div>
-        <.sender_user_image user={@sender} />
+        <.sender_user_image
+          state={
+            if @sender.profile_photo && is_atom(@sender.profile_photo.state) do
+              @sender.profile_photo.state
+            else
+              make_sure_photo_state_is_atom(@sender.profile_photo)
+            end
+          }
+          user={@sender}
+        />
       </div>
     <% end %>
     """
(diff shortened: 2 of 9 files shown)


lib/animina_web/components/bookmark/bookmark_component.ex
```diff
@@ -55,6 +55,17 @@ defmodule AniminaWeb.BookmarkComponent do
     end
   end
 
+  # we display the state as an atom in the photo struct, but we need to make sure it is an atom
```
wintermeyer, line 58: When I read this I ask myself WHY do we need to make sure if we already save it as an atom? Please add that this is probably a bug in Ash. Bonus points for adding a link to the bug report you created for Ash.

lib/animina_web/components/bookmark/bookmark_component.ex
```diff
@@ -65,6 +76,13 @@ defmodule AniminaWeb.BookmarkComponent do
         language={@language}
         reason={@reason}
         current_user={@current_user}
+        state={
+          if @current_user.profile_photo && is_atom(@current_user.profile_photo.state) do
```
wintermeyer, line 80: Don't we have a logic problem here? What if there is no photo? Than make_sure_photo_state_is_atom is called. Is that correct?

lib/animina_web/components/chat/chat_components.ex
```diff
@@ -115,7 +115,16 @@ defmodule AniminaWeb.ChatComponents do
             </div>
           </div>
         </div>
-        <.sender_user_image user={@sender} />
+        <.sender_user_image
+          state={
+            if @sender.profile_photo && is_atom(@sender.profile_photo.state) do
```
wintermeyer, line 120: Same potential logic problem if there is no `profile_photo`.

lib/animina_web/components/bookmark/bookmark_component.ex
```diff
@@ -55,6 +55,17 @@ defmodule AniminaWeb.BookmarkComponent do
...
   end
 
+  # we display the state as an atom in the photo struct, but we need to make sure it is an atom
+  # as we are using it for pattern matching
+
+  defp make_sure_photo_state_is_atom(nil) do
+    ""
+  end
+
+  defp make_sure_photo_state_is_atom(photo) do
```
wintermeyer, line 65: What happens if this already is an atom? Or are we sure that this can not happen?

lib/animina_web/components/chat/chat_components.ex
```diff
@@ -125,7 +134,17 @@ defmodule AniminaWeb.ChatComponents do
     ~H"""
     <%= if @sender.id != @message.sender_id do %>
       <div class="flex justify-start gap-4  item-start text-black">
-        <.receiver_user_image user={@receiver} current_user={@sender} />
+        <.receiver_user_image
+          state={
+            if @receiver.profile_photo && is_atom(@receiver.profile_photo.state) do
```
wintermeyer, line 139: This cries out loud to become some sort of helper function. DRY!

As fare as I understand it we are dealing with a potential Ash bug here. There will come a time when that bug is fixed. It is so much easier than to just fix one helper function than to walk through all the other code.

lib/animina_web/components/chat/chat_components.ex
```diff
@@ -321,4 +340,12 @@ defmodule AniminaWeb.ChatComponents do
   def format_time(time) do
     NaiveDateTime.from_erl!({{2000, 1, 1}, Time.to_erl(time)}) |> Timex.format!("{h12}:{0m} {am}")
   end
+
+  defp make_sure_photo_state_is_atom(nil) do
```
wintermeyer, line 344: DRY

wintermeyer (approved):
@MICHAELMUNAVU83 I approve this PR to fix the problem in production. But please do create a new issue to DRY this code and do that.

lib/animina_web/components/bookmark/bookmark_component.ex
MICHAELMUNAVU83: We have this is_atom(@sender.profile_photo.state)

lib/animina_web/components/chat/chat_components.ex
MICHAELMUNAVU83: Very True , I will have a seperate issue for this

lib/animina_web/components/bookmark/bookmark_component.ex
wintermeyer: > We have this is_atom(@sender.profile_photo.state)

So we check it in two different places? Wouldn't it be better to do this in one place/area?

MICHAELMUNAVU83: It would , I will refactor this code , already an issue for it

Issue #1139, opened 2024-11-12
Change the handling of deleting the last story


wintermeyer:
#1134 displayed a potential problem with the current logic.

I wanted to be sure that nor orphaned profiles without a single story are in the system but this leads to this problem. The only solution which I can come up with is that we allow users to delete even their about_you story (which is THE one story that should always be there) but auto change their visibility to hibernate (is that the name we use?) but only after we told the user that this is the consequence of remove the about_you story.

GUI wise we have to display it to a user when he/she is in hibernation. On every screen. In their face! And we have to make it easy to get back to normal mode again.

Please propose a better solution in this issue if you have an idea.

MICHAELMUNAVU83:
- [ ] Create a banner to be shown to the user all through the system if they are hibernated
- [ ] Ensure if a user hibernates themselevs in the profile visibility page , they can go back to normal if and only if they meet requirements eg profile photo and about me story
- [ ] Ensure a user can delete a profile photo for their profile  , we show them a modal informing them of consequences then they can delete , we need to now make the user hibernated.

PR #1149, opened 2024-11-14
Added Banner to be seen for hibernated users

MICHAELMUNAVU83:
@wintermeyer   , we have a banner now being seen if a user is hibernate  ,kindly advise if this is "IN your face" Enough for a user
Screenshot 2024-11-14 at 15 58 53

Screenshot 2024-11-14 at 15 49 14

### DIFF
--- a/lib/animina_web/components/top_navigation_components.ex
+++ b/lib/animina_web/components/top_navigation_components.ex
@@ -431,6 +431,21 @@ defmodule AniminaWeb.TopNavigationCompontents do
     """
   end
 
+  def hibernated_user_banner(assigns) do
+    ~H"""
+    <div
+      :if={@current_user && make_sure_user_state_is_atom(@current_user) == :hibernate}
+      class="mx-4 mb-4 p-2 text-center rounded-md  bg-blue-500 text-white"
+    >
+      <%= with_locale(@language, fn -> %>
+        <%= gettext(
+          "Your Account is Hibernated , This means users can't see your profile or send you messages. You can activate your account by ensuring you have a profile picture and an about me story with an image"
+        ) %>
+      <% end) %>
+    </div>
+    """
+  end
+
   defp get_photo_state_styling(:error) do
     "bg-red-500 text-white w-2 h-2 rounded-full"
   end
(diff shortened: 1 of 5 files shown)


lib/animina_web/components/top_navigation_components.ex
```diff
@@ -431,6 +431,21 @@ defmodule AniminaWeb.TopNavigationCompontents do
...
 
+  def hibernated_user_banner(assigns) do
+    ~H"""
+    <div
+      :if={@current_user && make_sure_user_state_is_atom(@current_user) == :hibernate}
+      class="mx-4 mb-4 p-2 text-center rounded-md  bg-blue-500 text-white"
+    >
+      <%= with_locale(@language, fn -> %>
+        <%= gettext(
+          "Your Account is Hibernated , This means users can't see your profile or send you messages. You can activate your account by ensuring you have a profile picture and an about me story with an image"
```
wintermeyer, line 442: That text assumes that the profile picture is missing (it doesn't but kind of it does) which doesn't make sense. A user might just hibernate his account for what ever reason. The text should be "This account is hibernated. Other users can't see it or send messages to you. Visit the `Change Visibility` page to activate it again." The "Change Visibility" part should be a link to that page.

On THAT page you have to make sure that one can not activate the account if there is not profile picture or about me story.

lib/animina_web/components/top_navigation_components.ex
```diff
@@ -511,6 +526,16 @@ defmodule AniminaWeb.TopNavigationCompontents do
     String.to_atom(photo.state)
   end
 
+  # We Have This checker so that incase we receive the state as a string , we
+  # still convert it to an atom
+  defp make_sure_user_state_is_atom(user) do
```
wintermeyer, line 531: Do we have to do this EVERYWHERE now? If yes, it would make more sense to not use the atom but always to use the string. Have you ask in the Elixirforum about this behaviour?

wintermeyer (changes requested):
I like the position of the banner. It is in your face.

lib/animina_web/components/top_navigation_components.ex
MICHAELMUNAVU83: I can have the first bit done , for the second one , I have a seperate issue for it here  https://github.com/animina-dating/animina/issues/1139#issuecomment-2476206925

lib/animina_web/components/top_navigation_components.ex
MICHAELMUNAVU83: Yes , You can follow the converssation here  https://elixirforum.com/t/using-database-values-stored-as-atoms-in-html-in-ash-framework/67437

PR #1154, opened 2024-11-16
Improve German translation

wintermeyer:


wintermeyer:
@briankariuki @MICHAELMUNAVU83 Can you guys help me to find the problem here? I have changed ONE word in the translation file. The test fails. What do I have to do differently?

briankariuki:
Rerun the workflow again. I've seen its a db connection error that caused the test to fail

PR #1160, opened 2024-11-20
Remove unhelpful dropdowns on the email verification page  , closes  #1124

### LINKED ISSUE #1124: Unhelpful pull down navigation entries during registration
During the registration process we display menus entries which do not make any sense for that point of the user journey. Not only should those entries not be displayed but also those URLs not accessible for users who don't have a full account yet.

Right now it is easy to stop the registration process by clicking on one of those entries and then the user is lost.

Bildschirmfoto 2024-11-05 um 18 17 53

MICHAELMUNAVU83:
We now have the email verificatio page having only 2 options , Log out or delete account.
Screenshot 2024-11-20 at 08 06 57

### DIFF
--- a/lib/animina_web/components/top_navigation_components.ex
+++ b/lib/animina_web/components/top_navigation_components.ex
@@ -264,9 +264,15 @@ defmodule AniminaWeb.TopNavigationCompontents do
               <% end) %>
             </.link>
           <% else %>
-            <.link class="px-2 w-[150px] text-center" navigate="/auth/user/sign-out">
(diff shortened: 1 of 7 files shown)


wintermeyer:
It is ensured that the other URLs are not available for a black hat hacker in case he knows them and can enter them in the browser? I don't want him to bypass us.

MICHAELMUNAVU83:
@wintermeyer  , yes , when I try to access a page and I have not confirmed my account , you are redirected back and shown the following flash

Screenshot 2024-11-20 at 10 14 17

wintermeyer (approved):

PR #1167, opened 2024-11-24
Fixed Credit Count Bug

MICHAELMUNAVU83:
- We had a bug where the credit points of a user and profile were not being shown correctly on the sidebar and the user profile page as such, we now calculate this value correctly.

https://github.com/user-attachments/assets/8668c92f-8b36-448c-9944-b75f491df3c2

### DIFF
--- a/lib/animina/genservers/profileview_credits.ex
+++ b/lib/animina/genservers/profileview_credits.ex
@@ -1,5 +1,6 @@
 defmodule Animina.GenServers.ProfileViewCredits do
   use GenServer
+  alias Animina.Accounts.User
   alias Phoenix.PubSub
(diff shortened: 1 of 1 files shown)


wintermeyer (approved):
I approve the code but do wonder if we can increase the speed by using FastUser here or even create a new function just for this. AFAIK now we load not just the gorilla but the whole jungle too.

@briankariuki or @MICHAELMUNAVU83  Please have a look at this and improve in case it as bad as I think it is.

MICHAELMUNAVU83:
Yeah , took some inspiration from some of @briankariuki  code for fast user , a PR is coming

PR #1170, opened 2024-11-26
add kamal config, closes #1169, closes #1173

### LINKED ISSUE #1169: Add Kamal configuration

### LINKED ISSUE #1173: Add a demo environment
We have testing, development and production. For a future demo system we need a new `demo` environment. Please add such a configuration. That configuration must use it's own database.

briankariuki:


lib/animina_web/plugs/health_check.ex
```diff
@@ -0,0 +1,19 @@
...
+  A Plug to return a health check on `/up`
+  """
+
+  import Plug.Conn
+
+  @behaviour Plug
+
+  def init(opts), do: opts
+
+  def call(%{path_info: ["up"]} = conn, _opts) do
```
wintermeyer, line 12: Did you add `up` on the blacklist for usernames? If not: Please do.

wintermeyer (changes requested):
Please add a `demo` environment. See https://github.com/animina-dating/animina/issues/1173

Please double check that we can run both environments on the same server without them interfering with each other. I don't want the `demo` environment deleting images from `production` and vice versa. Same with the database.

And please send @wintermeyer an email which outlines the move from the current production to the new system. Does anything have to be moved for that on the production system?

lib/animina_web/plugs/health_check.ex
briankariuki: I've added it to the list of bad usernames

briankariuki:
> Please double check that we can run both environments on the same server without them interfering with each other. I don't want the `demo` environment deleting images from `production` and vice versa. Same with the database.
> 

So long as each instance is mapped to its own database, port and file volume, there will be no issues.

wintermeyer:
@briankariuki The tests failed. Please check.

wintermeyer (approved):

PR #1171, opened 2024-11-26
Improved Query to get credit points for a user

MICHAELMUNAVU83:
Fixed Query for getting Credit Points for a user to ensure that is all we get and not the "Whole Jungle"

### DIFF
--- a/lib/animina/accounts/resources/fast_user.ex
+++ b/lib/animina/accounts/resources/fast_user.ex
@@ -28,6 +28,7 @@ defmodule Animina.Accounts.FastUser do
   code_interface do
     domain Animina.Accounts
     define :by_id_email_or_username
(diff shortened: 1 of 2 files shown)


lib/animina/accounts/resources/fast_user.ex
```diff
@@ -332,6 +374,46 @@ defmodule Animina.Accounts.FastUser do
       }
   end
 
+  defp build_query_for_credit_points(
```
wintermeyer, line 377: Why is this so complicated? Please add a comment to functions which are so complex why it has to be and what they are doing.

wintermeyer (approved):
@MICHAELMUNAVU83 Let's discuss this in our next call.

@briankariuki  Is there a better way to add this to FastUser?

PR #1176, opened 2024-11-29
Added a new route and template for the first 4 fields

MICHAELMUNAVU83:
- I added "beta" as a bad username and created a route "/beta" for the new registation workflow , it has the 4 initial fields we use ,Date of Birth , Zip Code , Gender and  Height  .
Screenshot 2024-11-29 at 08 53 45


lib/animina/validations/bad_username.ex
```diff
@@ -19,7 +19,7 @@ defmodule Animina.Validations.BadUsername do
     username =
       Ash.Changeset.get_attribute(changeset, opts[:attribute])
 
-    bad_usernames = ["my", "current_user", "profile"]
+    bad_usernames = ["my", "current_user", "profile", "beta"]
```
wintermeyer: + ["test", "new", "alpha", "next", "Stefan", "Stephan", "Aurelius", "Juna", "Miri", "Miriam", "aw", "jw", "sw"]

lib/animina/validations/bad_username.ex
```diff
@@ -19,7 +19,7 @@ defmodule Animina.Validations.BadUsername do
...
 
     if username &&
```
wintermeyer: While we are at it: Please add a validation, that a username has to have at least 2 characters.

wintermeyer (changes requested):
Bildschirmfoto 2024-11-29 um 09 18 21

## Thoughts

- For what ever reason the birthdate field is a constant source of trouble. Our users have problems to enter dates which are > 20 years ago. It seems to be easier for them to just enter a date as a text. We have to convert that into a date. And we have to make sure that the input makes sense. Input examples: "20.12.1970", "20.12.70", "20.12.01", "20.12.2000". What doesn't work: "20.12." (that would through a validation error).
- Let us start with a `"<h2>Your future partner</h2>"` section which asks for the gender. I know that I wanted to start with the user's gender first. But this way it makes more sense. After that we do a `"<h2>About you</h2>"` section where we ask the data from the user.

MICHAELMUNAVU83:
I also think we need to remove zip codes , or do Germans know exact zipcodes , If a zipcode is 56083 , that means I have to type in all 5 numbers to see a result , what are your thoughts?

wintermeyer:
> I also think we need to remove zip codes , or do Germans know exact zipcodes , If a zipcode is 56083 , that means I have to type in all 5 numbers to see a result , what are your thoughts?

Good thinking! But Germans do know their zip code. And yes, the user has to enter all 5 digits.

MICHAELMUNAVU83:
@wintermeyer  , one question , do you want it to be a 2 step page that is.
Step 1.
Select Gender and the zip code and see the number of potential partners , then hit proceed to go to the next stage/page or have them all in one oage , with a title under each so we have

1. Future Partner Information
2. Info about you
3. Then now the whole data page with email , username etc

wintermeyer:
> @wintermeyer , one question , do you want it to be a 2 step page that is. Step 1. Select Gender and the zip code and see the number of potential partners , then hit proceed to go to the next stage/page or have them all in one oage , with a title under each so we have
> 
> 1. Future Partner Information
> 2. Info about you
> 3. Then now the whole data page with email , username etc

One page with all the information until the flags but without the user and password info. Then a new page with the user and password thing. PIN. Then flags. Then about me story.

MICHAELMUNAVU83:
@wintermeyer  kindly confirm if this layout is now okay

Screenshot 2024-11-30 at 09 47 36

MICHAELMUNAVU83:
And also does this new layout mean that if someone selects a partners gender as male , we automatically have their gender as female and so on?

wintermeyer:
> @wintermeyer kindly confirm if this layout is now okay

The basic layout is ok. But the fields are not. Please ask youself "What information do we need to narrow down the potential partner?"  Those are the fields we are asking for on this very first page. Not `username` because that doesn't narrow down the search. Please add the things which narrow down the search from the second page to this page too.

We only ask for the birthdate and the height because that helps us to refill other fields automatically. Therefor it helps us to narrow down the search. But the `occupation` doesn't help us here.

Bildschirmfoto 2024-11-30 um 08 49 05

wintermeyer:
> And also does this new layout mean that if someone selects a partners gender as male , we automatically have their gender as female and so on?

Yes, we do use this information to make a guess about the gender of the user. But just to prefill that form field. The user can override it.

MICHAELMUNAVU83:
Screenshot 2024-11-30 at 11 29 09


I have removed the other fields , we now have it as such @wintermeyer

wintermeyer:
> I have removed the other fields , we now have it as such @wintermeyer

Very well. Now add these field to the same form.

Bildschirmfoto 2024-11-30 um 10 20 26

MICHAELMUNAVU83:
Should this be under the "Future Partner Information" section or  "Information about you" part @wintermeyer

wintermeyer:
> Should this be under the "Future Partner Information" section or "Information about you" part @wintermeyer

For now: Add an other "Future Partner Information" section below the current "Information about you" one. We'll figure out the design later. I want a workflow which narrows down the list of potential partners. A sort of funnel.

MICHAELMUNAVU83:
Like this ? @wintermeyer 

Screenshot 2024-11-30 at 12 36 49
Screenshot 2024-11-30 at 12 36 43

wintermeyer:
> Like this ? @wintermeyer

Yes! But please reorder to this order:

- Height
- Birthdate
- ZipCode

Why? That gives us a couple of more milliseconds to pre-fill the values below. So when somebody enters the `height` we calculate the default values for the min and max partner height. Same with the age. Of course the user can override it but a lot of people will not and for them it is a real help and usability improvement.

MICHAELMUNAVU83:
That does make sense 
Screenshot 2024-11-30 at 12 47 43


Also advice on how we should have defaults for heights , can we do +- 20 centimeters and for age +- 5 years?

wintermeyer:
> Also advice on how we should have defaults for heights , can we do +- 20 centimeters and for age +- 5 years?

Yes, those values make sense. And as always: Should be config variables which an admin can change in config.exs, dev.exs, test.exs or prod.exs and not manually in the hardcodes LiveView code.

MICHAELMUNAVU83:
Okay , we can merge this , then I can work on the potential partner algorithm

wintermeyer (approved):

PR #1177, opened 2024-12-02
Created Algorithm for Potential Partners in Beta Register Page

MICHAELMUNAVU83:
https://github.com/user-attachments/assets/aab2fb0d-b0ba-4754-91bb-1481f691952a

@wintermeyer  , finished up on the algorithm which works as such

**1. Gender**

The first selection is gender , by default, the partner gender is set to male , we then filter all records to only bring male ones

**2. Age**

If a user selects their age eg 20 , we add and minus 10 to get the maximum and minimum age if these are not selected , if they are , we use the maximum and minimum age that the user explicitly selects and filter users against it. We have the difference to add set in config.

**3. Height**

 If a user selects their height eg 160 , we add and minus 10 to get the maximum and minimum partner height if these are not selected , if they are , we use the maximum and minimum height that the user explicitly selects and filter users against it. We have the difference to add set in config.

**4. Search Range and Zip Code**

If a user selects a zip code, but no search range , we use a high value as the search range , in this case `10000` , we have this set in the config , if the user selects the search range explicitly though , we use this value.

The next step is to add validations to this page and move to the next page where a user completes their registration.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -31,6 +31,27 @@ config :animina, :ai_message_help_price, 20
 # Configures whether to have ai automatically respond to messages
 config :animina, :autoreply_messages_with_ai, false
 
+# Configures default maximum age for potential partners
+config :animina, :default_potential_partner_maximum_age, 100
+
+# Configures default minimum age for potential partners
+config :animina, :default_potential_partner_minimum_age, 18
+
+# confugures the difference in age if no maximum or minimum age is set
+config :animina, :default_potential_partner_age_difference, 10
+
+# Configures the maximum height for potential partners
+config :animina, :default_potential_partner_maximum_height, 300
+
+# Configures the minimum height for potential partners
+config :animina, :default_potential_partner_minimum_height, 0
+
+# Configures the maximum height difference for potential partners
+config :animina, :default_potential_partner_height_difference, 10
+
+# Configures the default search range for potential partners
+config :animina, :default_potential_partner_search_range, 10_000
+
 config :animina,
   ecto_repos: [Animina.Repo],
   generators: [timestamp_type: :utc_datetime]
--- a/config/config.exs
+++ b/config/config.exs
@@ -38,19 +38,19 @@ config :animina, :default_potential_partner_maximum_age, 100
 config :animina, :default_potential_partner_minimum_age, 18
 
 # confugures the difference in age if no maximum or minimum age is set
-config :animina, :default_potential_partner_age_difference, 10
+config :animina, :default_partner_age_offset, 10
 
 # Configures the maximum height for potential partners
-config :animina, :default_potential_partner_maximum_height, 300
+config :animina, :default_potential_partner_maximum_height, 225
 
 # Configures the minimum height for potential partners
 config :animina, :default_potential_partner_minimum_height, 0
 
 # Configures the maximum height difference for potential partners
-config :animina, :default_potential_partner_height_difference, 10
+config :animina, :default_partner_height_offset, 10
 
-# Configures the default search range for potential partners
-config :animina, :default_potential_partner_search_range, 10_000
+# Configures the default search range for potential partners in km
+config :animina, :default_potential_partner_search_range_in_km, 10_000
 
 config :animina,
   ecto_repos: [Animina.Repo],
--- a/lib/animina_web/components/beta_registration_components.ex
+++ b/lib/animina_web/components/beta_registration_components.ex
@@ -28,7 +28,7 @@ defmodule AniminaWeb.BetaRegistrationComponents do
 
       <h2 class="text-2xl mt-3 font-semibold dark:text-white">
         <%= with_locale(@language, fn -> %>
-          <%= gettext("Current Potential Partners:") %> 123
+          <%= gettext("Current Potential Partners:") %> <%= @number_of_potential_partners %>
         <% end) %>
       </h2>
 
@@ -37,7 +37,7 @@ defmodule AniminaWeb.BetaRegistrationComponents do
         id="beta_user_registration_form"
         for={@form}
         class="space-y-6 mt-6 group "
-        phx-change="validate"
+        phx-change="validate_and_filter_potential_partners"
         phx-submit="submit"
       >
         <h2 class="text-2xl mt-3 font-semibold dark:text-white">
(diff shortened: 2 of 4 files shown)


config/config.exs
```diff
@@ -31,6 +31,27 @@ config :animina, :ai_message_help_price, 20
...
 config :animina, :autoreply_messages_with_ai, false
 
+# Configures default maximum age for potential partners
+config :animina, :default_potential_partner_maximum_age, 100
+
+# Configures default minimum age for potential partners
+config :animina, :default_potential_partner_minimum_age, 18
+
+# confugures the difference in age if no maximum or minimum age is set
+config :animina, :default_potential_partner_age_difference, 10
```
wintermeyer: What does this mean? If I am 50 years old will that mean that the range is from 40 - 60 or is it from 45 - 55? Either improve the actual code or add a better comment to it.

config/config.exs
```diff
@@ -31,6 +31,27 @@ config :animina, :ai_message_help_price, 20
...
+
+# Configures the maximum height for potential partners
+config :animina, :default_potential_partner_maximum_height, 300
```
wintermeyer: 300 as a default doesn't make any sense. Make it 120 please.

PS: There will come the day when I get an email from a user who has a height of 121 cm. I am so sure! ;-)

config/config.exs
```diff
@@ -31,6 +31,27 @@ config :animina, :ai_message_help_price, 20
...
+
+# Configures the minimum height for potential partners
+config :animina, :default_potential_partner_minimum_height, 0
```
wintermeyer, line 47: Doesn't make sense either. Plus what is this variable about? Is this the default value for the form or for some sort of validation?

config/config.exs
```diff
@@ -31,6 +31,27 @@ config :animina, :ai_message_help_price, 20
...
+
+# Configures the maximum height difference for potential partners
+config :animina, :default_potential_partner_height_difference, 10
+
+# Configures the default search range for potential partners
+config :animina, :default_potential_partner_search_range, 10_000
```
wintermeyer: What is 10_000 here? km? cm?

Please assume that the code is read by somebody who only has this piece of code in front of him and who wants to understand what you are doing.

lib/animina_web/components/beta_registration_components.ex
```diff
@@ -319,28 +319,32 @@ defmodule AniminaWeb.BetaRegistrationComponents do
...
-              )
+              ),
+          autofocus: true
         ) %>
 
-        <.error :for={msg <- get_field_errors(@f[:minimum_partner_age], :minimum_partner_age)}>
+        <.error :for={msg <- get_field_errors(@f[:minimum_partner_height], :minimum_partner_height)}>
           <%= with_locale(@language, fn -> %>
-            <%= gettext("Minimum age") <> " " <> msg %>
+            <%= gettext("Minimum height") <> " " <> msg %>
```
wintermeyer, line 347: Please ALWAYS(!) use something like `gettext("Hello, %{name}!", name: "Meg")` for this kind of thing. All languages are different. There is no way this order will not change for all languages.

lib/animina_web/potential_partner.ex
```diff
@@ -52,6 +52,27 @@ defmodule AniminaWeb.PotentialPartner do
...
 
+  @doc """
+  We use this to fetch potential partners for the user during registration.
+  """
+
+  def potential_partners_on_registration(user) do
+    User
+    |> Ash.Query.for_read(:read)
+    |> Ash.Query.sort(Ash.Sort.expr_sort(fragment("RANDOM()")))
+    |> beta_registration_partner_gender_query(user)
```
wintermeyer: With this kind of naming we run into the "What do we do when this code moves from beta to normal?" are we doing a search and replace? Is that save? Would it be maybe better to just assume that this is not beta since we are working within this context?

lib/animina_web/potential_partner.ex
```diff
@@ -213,4 +356,16 @@ defmodule AniminaWeb.PotentialPartner do
         []
     end
   end
+
+  def convert_to_date(date_string) do
+    {:ok, date} = Date.from_iso8601(date_string)
+    date
+  end
+
+  def calculate_age(birthdate) do
```
wintermeyer, line 365: This can lead to some edge cases. Please use something like this:

```elixir
def calculate_age(birthdate) do
  today = Date.utc_today()
  years = today.year - birthdate.year

  if Date.compare(today, ~D[#{birthdate.year + years}-#{birthdate.month}-#{birthdate.day}]) == :lt do
    years - 1
  else
    years
  end
end
```

And make sure that `birthdate` is not in the future. Example:

```elixir
def calculate_age(birthdate) when birthdate > Date.utc_today(), do: {:error, "Birthdate is in the future"}
```

wintermeyer (changes requested):

config/config.exs
MICHAELMUNAVU83: It means range is from 40 and 60 , Kindly advise on the best naming to have in this case @wintermeyer

config/config.exs
MICHAELMUNAVU83: In this case we want a value so huge that will make sure we do not leave out any users , we also use this value as the default maximum height in the query , this allows us to select all users below 300 cm , I think it would be best to have a very high value for this case , thoughts? @wintermeyer

config/config.exs
MICHAELMUNAVU83: No , this value is the one we use when the page loads or the user has not set any height , we want to have the lowest number possible as a minimum height and the highest number as the maximum height so by default we show as many users as possible

lib/animina_web/potential_partner.ex
MICHAELMUNAVU83: The next PR was to handle validations that are already there in the user resource for this page.

config/config.exs
wintermeyer: A quick google research let to the number 251. That seems to be the height human right now. Let's go with 225 cm since the heightest German right now seems to be 224.

config/config.exs
wintermeyer: Let's call it `default_partner_height_offset`

Not perfect but good for now.

config/config.exs
MICHAELMUNAVU83: This is Km , I have fixed this

lib/animina_web/potential_partner.ex
MICHAELMUNAVU83: I have changed the naming of this

MICHAELMUNAVU83:
@wintermeyer  , kindly have a  look

PR #1184, opened 2024-12-05
Change file uploads path, closes #1183

### LINKED ISSUE #1183: Add file uploads volume

briankariuki:
@wintermeyer As discussed via email, you will need to move the current content in `priv/static/uploads` to a folder on the host system. You can then set this folder path as the `*_FILE_VOLUME_PATH` in the environment variables. Inside the docker container the volume will be mounted on the root at `/uploads`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,9 @@ RUN mix local.hex --force && \
 # set build ENV
 ENV MIX_ENV="prod"
 
+# allow builds on apple silicon macs
+ENV ERL_AFLAGS="+JMsingle true"
+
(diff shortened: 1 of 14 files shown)


wintermeyer:
@briankariuki Will this only work for the new deployment or can I use the same configuration for the old deployment?

briankariuki:
> @briankariuki Will this only work for the new deployment or can I use the same configuration for the old deployment?

Updated with comments. It will work with current deployment. But you will need to move the files before starting the docker container

wintermeyer (approved):

